PR #93, opened 2023-05-05
[DECO-1115] Add local implementation for `dbutils.widgets`

kartikgupta-db:
## Changes
* Added a new install group (`pip install 'databricks-sdk[notebook]'`). This allows us to safely pin ipywidgets for local installs. DBR can safely continue using `pip install databricks-sdk` or directly using the default build from master without conflicting dependencies. 
* OSS implementation of widgets is imported only on first use (possible only through OSS implementation of dbutils - `RemoteDbutils`). 
* Add a wrapper for ipywidgets to enable interactive widgets when in interactive **IPython** notebooks. 

https://user-images.githubusercontent.com/88345179/236443693-1c804107-ba21-4296-ba40-2b1e8e062d16.mov

* Add default widgets implementation that returns a default value, when not in an interactive environment.

https://user-images.githubusercontent.com/88345179/236443729-51185404-4d28-49c6-ade0-a665e154e092.mov

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied


kartikgupta-db:
TODO: tests

databricks/sdk/_widgets/IPyWidgetUtils.py
```diff
@@ -0,0 +1,89 @@
...
+
+        raise ValueError("The returned value has invalid type (" + type(value) + ").")
+
+
+class IPyWidgetUtil(_WidgetUtils):
+
+    def __init__(self) -> None:
+        self.__widgets: typing.Dict[str, DbUtilsWidget] = {}
+
+    def __register(self, name: str, widget: ValueWidget, label: typing.Optional[str] = None):
```
nfx, line 44:
```suggestion
    def _register(self, name: str, widget: ValueWidget, label: typing.Optional[str] = None):
```
one underscore is enough :) I don't think a lot of people will be extending it.

databricks/sdk/_widgets/WidgetUtils.py
```diff
@@ -0,0 +1,80 @@
+import typing
+from abc import ABC, abstractmethod
+
+
+class _WidgetUtils(ABC):
+
+    @property
```
nfx, line 7: have consistent sequencing for abstract and non-abstract methods

databricks/sdk/_widgets/__init__.py
```diff
@@ -0,0 +1,15 @@
+import os
+
+try:
+    from IPython.core.getipython import get_ipython
```
nfx, line 4: what about working in DBR?

nfx (changes requested):

tests/test_dbutils.py
```diff
@@ -188,14 +188,14 @@ def test_any_proxy(dbutils_proxy):
     command = ('\n'
                '        import json\n'
                '        (args, kwargs) = json.loads(\'[["a"], {}]\')\n'
-               '        result = dbutils.widgets.getParameter(*args, **kwargs)\n'
+               '        result = dbutils.notebook.exit(*args, **kwargs)\n'
```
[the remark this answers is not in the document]
kartikgupta-db: This tests is failing because 
1. getParameter is not an actual widget function
2. widgets no longer use proxy

databricks/sdk/_widgets/__init__.py
kartikgupta-db: This works in dbr

databricks/sdk/_widgets/IPyWidgetUtils.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @property
+    def _widgetNames(self) -> typing.List[str]:
+        return list(self.__widgets.keys())
+
+    def _get(self, name: str) -> str:
+        return self.__widgets[name].value
+
+    def _remove(self, name: str):
+        del self.__widgets[name]
```
xiaochen-db, line 85:
```suggestion
    def _remove(self, name: str):
        if name in self.__widgets:
            self.__widgets[name].close()
            del self.__widgets[name]
```

databricks/sdk/_widgets/__init__.py
kartikgupta-db: @xiaochen-db we do not use the this version (the oss version) of dbutils within dbr at all (https://github.com/databricks/databricks-sdk-py/pull/96). So we shouldn't need to override widgets specifically. Same for ipywidget versioning. The oss implementation is activated only when outside dbr (when a default global implementation of dbutils is not available).

databricks/sdk/_widgets/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+import warnings
+from abc import ABC, abstractmethod
+
+
+class WidgetUtils(ABC):
+
+    @property
+    @abstractmethod
+    def _widget_names(self) -> typing.List[str]:
+        pass
```
mgyucht, line 12: Maybe remove this if it is not being used

databricks/sdk/_widgets/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+    # We only use ipywidgets if we are in a notebook interactive shell otherwise we raise error,
+    # to fallback to using default_widgets. Also, users WILL have IPython in their notebooks (jupyter),
+    # because we DO NOT SUPPORT any other notebook backends, and hence fallback to default_widgets.
+    from IPython.core.getipython import get_ipython
+
+    # Iterate over the mro of the current ipython instance, to find ZMQInteractiveShell (jupyter). When used
+    # from REPL or file, this check will fail, since the mro only contains TerminalInteractiveShell.
+    if len(list(filter(lambda i: i.__name__ == 'ZMQInteractiveShell', get_ipython().__class__.__mro__))) == 0:
+        logging.debug("Not in an interactive notebook. Skipping ipywidgets implementation for dbutils.")
+        raise EnvironmentError("Not in an interactive notebook.")
```
mgyucht, line 52: Not a blocker for this change, but have we thought about how widgets would work for users running programs using the Databricks SDK locally? Current fallback implementation uses the default value only, but perhaps that implementation could look for an environment variable named `DB_WIDGET_<NAME>` to allow automated use-cases to set widget values.

databricks/sdk/_widgets/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+
+    except ImportError as e:
+        # Since we are certain that we are in an interactive notebook, we can make assumptions about
+        # formatting and make the warning nicer for the user.
+        warnings.warn(
+            "\nTo use databricks widgets interactively in your notebook, please install databricks sdk using:\n"
+            "\tpip install 'databricks-sdk[notebook]'\n"
+            "Falling back to default_value_only implementation for databricks widgets.")
+        logging.debug(f"{e.msg}. Skipping ipywidgets implementation for dbutils.")
+        raise e
```
mgyucht, line 70: When would we actually hit this case? Are there Jupyter notebooks that don't have the ipython dependencies installed?

databricks/sdk/_widgets/__init__.py
kartikgupta-db: Let me think about this offline and we can make this a separate PR. This PR is already getting complicated.

databricks/sdk/_widgets/__init__.py
kartikgupta-db: `ipywidgets` is not installed by default in jupyter environments. That's when we hit this case. 

Also, we can hit this case, if any dependency is too old, or missing some features we depend on (eg Combobox from ipywidgets).

databricks/sdk/_widgets/IPyWidgetUtils.py
```diff
@@ -0,0 +1,89 @@
+import typing
+
+from IPython.core.display_functions import display
+from ipywidgets.widgets import (ValueWidget, Widget, widget_box,
```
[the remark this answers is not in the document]
kartikgupta-db: Users can access dbutils from 2 points
  a. `from databricks.sdk.runtime import *`: This always falls back to the dbr version of dbutils. The oss implementations do not come into effect. 
  b. `workspace_client.dbutils`: This always uses the open source version of dbutils. We are working on a fix so that this property also falls back to the dbr version of dbutils when possible. 
This means that any imports for ipywidgets and other deps do not take effect when on dbr. 

Further, I added a new install group (`databricks-sdk[notebook]`), which has certain pinned versions for widgets and ipython. Users are prompted to install it if we detect that users are in an interactive notebook environment, but by default this will not be installed on the cluster. These checks are also bypassed in the dbr in favour of dbr native dbutils.

mgyucht (approved):

databricks/sdk/_widgets/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+
+    # For import errors in IPyWidgetUtil, we provide a warning message, prompting users to install the
+    # correct installation group of the sdk.
+    try:
+        from .ipywidgets_utils import IPyWidgetUtil
+
+        widget_impl = IPyWidgetUtil
+        logging.debug("Using ipywidgets implementation for dbutils.")
+
+    except ImportError as e:
```
xiaochen-db, line 58: How can user end up into this? Not installing `ipywidgets`?

kartikgupta-db: Yes. Or if they are on the wrong version (which doesn't have some packages we depend on).

xiaochen-db (approved):

PR #96, opened 2023-05-05
[PROD-37308] Preserve original databricks.sdk.runtime for internal purposes

dby-tmwctw:
## Changes
We want to use `databricks-sdk` in DBR but currently installing it breaks existing `databricks.sdk.runtime` module we added. This PR fix it by using the internal packages if existed, and fall-back to OSS implementation if that does not exist.

## Tests
- Manual testing on E2 Dogfood

- [ ] `make test` run locally
- [x] `make fmt` applied


databricks/sdk/runtime/__init__.py
```diff
@@ -6,19 +6,42 @@ def _get_global_dbutils():
         return None
 
 
-dbutils = _get_global_dbutils()
+is_oss_implementation = True
```
nfx, line 9: i think a better name could be “local implementation”

databricks/sdk/runtime/__init__.py
```diff
@@ -6,19 +6,42 @@ def _get_global_dbutils():
...
-        # this assumes that all environment variables are set
-        dbutils = RemoteDbUtils()
-except NameError:
-    from databricks.sdk.dbutils import RemoteDbUtils
-    if not dbutils:
-        # this assumes that all environment variables are set
-        dbutils = RemoteDbUtils()
+    # OSS implementation
+    is_oss_implementation = True
+    dbutils = _get_global_dbutils()
```
nfx, line 30: Can you check if we can remove _get_global_dbutils function altogether? It’s meant to check if “dbutils” got injected into nonlocal namespace

databricks/sdk/runtime/__init__.py
```diff
@@ -6,19 +6,42 @@ def _get_global_dbutils():
...
+
+    try:
+        from .stub import *
```
nfx, line 33: Stub is only there for VSCODE extension to show Autocompletion with documentation. It has no implementation, as you see.

@kartikgupta-db do you think we can remove stub and keep VSCode extension functioning properly?

databricks/sdk/runtime/__init__.py
```diff
@@ -6,19 +6,42 @@ def _get_global_dbutils():
...
+    except ImportError:
+        from databricks.sdk.dbutils import RemoteDbUtils
+        if not dbutils:
+            # this assumes that all environment variables are set
+            dbutils = RemoteDbUtils()
+    except NameError:
+        from databricks.sdk.dbutils import RemoteDbUtils
```
nfx, line 40: Why do we need to import remote dbutils implementation twice?…

nfx (changes requested):
Few comments

databricks/sdk/runtime/__init__.py
kartikgupta-db: Not for now. We have started fully depending on the SDK for the stubs. So we need to keep them in.

databricks/sdk/runtime/__init__.py
kartikgupta-db: +1. here you are explicitly injecting dbutils into the globals namespace. So we don't really need  to see if dbutils are injected when `is_oss_implementation = True`

databricks/sdk/runtime/__init__.py
```diff
@@ -6,19 +6,42 @@ def _get_global_dbutils():
...
+        if not dbutils:
+            # this assumes that all environment variables are set
+            dbutils = RemoteDbUtils()
```
kartikgupta-db, line 43:
```suggestion
    except (ImportError, NameError):
        from databricks.sdk.dbutils import RemoteDbUtils
        if not dbutils:
            # this assumes that all environment variables are set
            dbutils = RemoteDbUtils()
```

databricks/sdk/runtime/__init__.py
nfx: @kartikgupta-db can we somehow separate the stubs even further within the SDK?

databricks/sdk/runtime/__init__.py
dby-tmwctw: @nfx This is the original OSS implementation, but the commit suggestion by @kartikgupta-db down seems to solve this issue.

databricks/sdk/runtime/__init__.py
dby-tmwctw: Makes sense, let me change the name.

databricks/sdk/runtime/__init__.py
dby-tmwctw: @nfx @kartikgupta-db Makes sense, may I know how to test whether it is OK to remove it though, is passing unit tests sufficient?

nfx (approved):
Looks good to me, let’s wait for @kartikgupta-db to approve from VSCode part

databricks/sdk/runtime/__init__.py
nfx: We don’t have unit tests for this module yet.

databricks/sdk/runtime/__init__.py
```diff
@@ -1,24 +1,33 @@
...
+    except (ImportError, NameError):
+        from databricks.sdk.dbutils import RemoteDbUtils
+
         # this assumes that all environment variables are set
         dbutils = RemoteDbUtils()
 
-__all__ = ['dbutils']
+__all__ = ['dbutils'] if is_local_implementation else [
+    "display", "displayHTML", "dbutils", "table", "sql", "udf", "getArgument", "sc", "sqlContext", "spark"
+]
```
xiaochen-db, line 33: nit: store this list as constant since it repeats twice

dby-tmwctw: Done.

nfx (approved):
lgtm

PR #97, opened 2023-05-08
Release v0.1.5

nfx:
* Pin version of `requests` to `<2.30.0`, so that we don't get `ImportError: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with LibreSSL 2.8.3`. See [this issue](https://github.com/psf/requests/issues/6432) for more information.


setup.py
```diff
@@ -10,7 +10,7 @@
       version=version_data['__version__'],
       packages=find_packages(exclude=["tests", "*tests.*", "*tests"]),
       python_requires=">=3.7",
-      install_requires=["requests>=2.28.1"],
+      install_requires=["requests<2.30.0"],
```
pietern, line 13: Please keep lower bound (I'm sure it's there for a good reason).

pietern (approved):

PR #110, opened 2023-05-15
Persist token acquired through 'external-browser' auth type

pietern:
## Changes

This cache is local to the Python SDK and keyed off the workspace host, the OAuth client ID, and the list of scopes to authorize for. The cache path is `~/.config/databricks-sdk-py/oauth`. Files saved to this directory are masked 0600.

It does not persist refreshes that happen during a session.

## Tests

Reproduce by running `./examples/local_browser_oauth.py` multiple times.

- [x] `make test` run locally
- [x] `make fmt` applied

### DIFF
--- a/databricks/sdk/oauth.py
+++ b/databricks/sdk/oauth.py
@@ -1,7 +1,9 @@
 import base64
 import functools
 import hashlib
+import json
 import logging
+import os
 import secrets
 import threading
 import urllib.parse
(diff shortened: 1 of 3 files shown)


databricks/sdk/oauth.py
```diff
@@ -384,3 +386,41 @@ def refresh(self) -> Token:
                               params,
                               use_params=self.use_params,
                               use_header=self.use_header)
+
+
+class TokenCache():
+    BASE_PATH = "~/.config/databricks-sdk-py/oauth"
```
nfx, line 392: Probably it's better to use `~/.databricks/oauth`

nfx (approved):
will make necessary renames later

PR #117, opened 2023-05-17
use shell for opening az cli

grusin-db:
## Changes

handle calling `az login` or alike cmd commands on windows 10 (and possibly other versions of windows). it seems that windows (and only windows) requires `shell=True` when executing sub processes. 

code checks if running on windows and only uses shell execution mode.

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied
- [x] tested additionally conditional code on windows machine, but outside of build system, just on sample copied code (build tests on windows machine are not working yet...).

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -224,7 +224,7 @@ class CliTokenSource(Refreshable):
 
     def refresh(self) -> Token:
         try:
-            out = subprocess.check_output(self._cmd, stderr=subprocess.STDOUT)
+            out = subprocess.check_output(self._cmd, stderr=subprocess.STDOUT, shell=True)
             it = json.loads(out.decode())
             expires_on = self._parse_expiry(it[self._expiry_field])
             return Token(access_token=it[self._access_token_field],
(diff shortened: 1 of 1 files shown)


databricks/sdk/core.py
```diff
@@ -224,7 +225,10 @@ def _parse_expiry(expiry: str) -> datetime:
 
     def refresh(self) -> Token:
         try:
-            out = subprocess.check_output(self._cmd, stderr=subprocess.STDOUT)
+            is_windows = sys.platform.startswith('win')
+            #windows requires shell=True to be able to execute 'az login' or other commands
```
nfx, line 229:
```suggestion
            # windows requires shell=True to be able to execute 'az login' or other commands
```

databricks/sdk/core.py
```diff
@@ -224,7 +225,10 @@ def _parse_expiry(expiry: str) -> datetime:
...
+            #cannot use shell=True all the time, as it breaks macOS
```
nfx, line 230:
```suggestion
            # cannot use shell=True all the time, as it breaks macOS
```

nfx (approved):

PR #118, opened 2023-05-17
Harden `w.clusters.ensure_cluster_running()`

nfx:
## Changes

- Retry on `OperationFailed` errors, same as in Java SDK

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


judahrand:
I think you need to handle the case where the cluster is already running.

PR #138, opened 2023-06-01
Relax `requests` version requirement

### LINKED ISSUE #136: Please loosen the dependency on `requests`
This package currently depends on `requests <2.29.0,>=2.28.1`, which is a very tight range and unfortunately has published vulnerabilities. Please widen the allowed range (I suggest to `requests <3,>=2.28.1` if forwards compatibility is important) so that this package does not force insecure dependent packages on its consumers.

judahrand:
## Changes

Closes #136 

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


setup.py
```diff
@@ -10,7 +10,7 @@
       version=version_data['__version__'],
       packages=find_packages(exclude=["tests", "*tests.*", "*tests"]),
       python_requires=">=3.7",
-      install_requires=["requests>=2.28.1,<2.29.0"],
+      install_requires=["requests>=2.28.1,<3"],
```
nfx, line 13: we cannot increase the upper bound, as some of our downstream packages didn't migrate yet to requests 2.29.0+, lower bound can be decreased if it passes the tests.

see https://github.com/databricks/databricks-sdk-py/commit/1ea8f3b43e2e2b40a022f9bc1a2f9167dd629c57 and https://github.com/psf/requests/issues/6432

nfx (changes requested):

setup.py
judahrand: > we cannot increase the upper bound, as some of our downstream packages didn't migrate yet to requests 2.29.0+

Surely, the downstream packages should be responsible for pinning requests?! This is a publicly available package which can't be used with a version of requests without public security vulnerabilities (https://github.com/psf/requests/blob/main/HISTORY.md#2310-2023-05-22)...

PeterJCLaw: @nfx am I correct in understanding that the issue with supporting requests 2.29+ is to do with a mismatch in available ssl modules and `urllib3` version 2.x? If so, perhaps this package could pin to `urllib3<2` (as the release notes for `requests` 2.30 suggest) as that would appear to avoid the issue? That would allow for `requests` to move forward to a version which has the security fix (i.e: 2.31) while keeping to the known-working version of `urllib3`.

I agree with @judahrand that such pinning really ought to be the responsibility of the downstream libraries/applications which actually depend on the particular versions of `requests`/`urllib3`, however if it's important to you that this package not admit `urllib3` related issues then perhaps my suggestion above presents a way forwards.

mgyucht: @PeterJCLaw @judahrand Sorry for the delay in addressing this ticket. As we saw during the original bump of urllib3, major version increases with backwards incompatibility has major risks for dependent software. We wanted to verify that at least our internal products wouldn't have yet another regression if we were to unpin this too soon. Especially with the release of 2.0.3, users who do not pin urllib3 in their setup.py or requirements.txt are likely to get a version that is compatible with their system's version of SSL. So, we're ready to merge this patch. This will be released in the next minor version (0.2.0).

PR #139, opened 2023-06-01
Databricks Metadata Service credential provider

kartikgupta-db:
## Changes
* Port databricks metadata service credential provider from go sdk.

## Tests
* unit tests

- [x] `make test` run locally
- [x] `make fmt` applied

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -448,7 +448,7 @@ class Config:
     debug_headers: bool = ConfigAttribute(env='DATABRICKS_DEBUG_HEADERS')
     rate_limit: int = ConfigAttribute(env='DATABRICKS_RATE_LIMIT')
     retry_timeout_seconds: int = ConfigAttribute()
-    metadata_service_url = ConfigAttribute(env='DATABRICKS_METADATA_SERVICE_URL', auth='metadata-service')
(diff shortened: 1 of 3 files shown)


databricks/sdk/core.py
```diff
@@ -312,6 +312,57 @@ def inner() -> Dict[str, str]:
     return inner
 
 
+class MetadataServiceTokenSource(Refreshable):
+    """ Obtain the token granted by Databricks Metadata Service """
+    MetadataServiceVersion = "1"
```
nfx, line 317: make them ALL_CAPS - this is the way how Python does constants

tests/test_auth.py
```diff
@@ -254,3 +259,76 @@ def test_config_auth_type_from_env(monkeypatch):
 
     assert cfg.auth_type == 'basic'
     assert cfg.host == 'https://x'
+
+
+def get_test_server(host: str, token: str, expires_after: int):
```
nfx, line 264:
```
# These are auto-generated tests for Unified Authentication
# In case of editing this file, make sure the change is propagated to all Databricks SDK codebases
```

nfx (changes requested):

nfx (approved):

PR #148, opened 2023-06-02
Consistently use the `upload(path, IO)` and `download(path) -> IO` across file-related operations

### LINKED ISSUE #104: Notebook export with "direct_download=True" fails
Setting the `direct_download` flag to True in the API returns an unencoded version of the notebook but throws a non-descript error today.

nfx:
## Changes

- added `w.workspace.upload` & `w.workspace.download`
- added `w.dbfs.upload` & `w.dbfs.download`
- added `w.files.upload` & `w.files.download`
- modified low-level client to work with raw streams and debug messages correctly

Fix #104

## Tests
new integration tests

### DIFF
--- a/.codegen/__init__.py.tmpl
+++ b/.codegen/__init__.py.tmpl
@@ -1,7 +1,7 @@
 import databricks.sdk.core as client
 import databricks.sdk.dbutils as dbutils
 
-from databricks.sdk.mixins.dbfs import DbfsExt
+from databricks.sdk.mixins.files import DbfsExt, FilesMixin
 from databricks.sdk.mixins.compute import ClustersExt
 from databricks.sdk.mixins.workspace import WorkspaceExt
 {{- range .Services}}
@@ -35,6 +35,7 @@ class WorkspaceClient:
         self.config = config
         self.dbutils = dbutils.RemoteDbUtils(self.config)
         self.api_client = client.ApiClient(self.config)
+        self.files = FilesMixin(self.api_client)
         {{- range .Services}}{{if not .IsAccounts}}
         self.{{.SnakeName}} = {{template "api" .}}(self.api_client){{end -}}{{end}}
 
--- a/.codegen/__init__.py.tmpl
+++ b/.codegen/__init__.py.tmpl
@@ -3,6 +3,7 @@ import databricks.sdk.dbutils as dbutils
 
 from databricks.sdk.mixins.dbfs import DbfsExt
 from databricks.sdk.mixins.compute import ClustersExt
+from databricks.sdk.mixins.workspace import WorkspaceExt
 {{- range .Services}}
 from databricks.sdk.service.{{.Package.Name}} import {{.PascalName}}API{{end}}
 
@@ -11,7 +12,7 @@ from databricks.sdk.service.{{.Package.Name}} import {{.PascalName}}API{{end}}
   "azure_client_id" "azure_tenant_id" "azure_environment" "auth_type" "cluster_id"}}
 
 {{- define "api" -}}
-  {{- $mixins := dict "ClustersAPI" "ClustersExt" "DbfsAPI" "DbfsExt" -}}
+  {{- $mixins := dict "ClustersAPI" "ClustersExt" "DbfsAPI" "DbfsExt" "WorkspaceAPI" "WorkspaceExt" -}}
   {{- $genApi := concat .PascalName "API" -}}
   {{- getOrDefault $mixins $genApi $genApi -}}
 {{- end -}}
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -825,26 +825,34 @@ class ApiClient:
             r.headers[k] = v
         return r
 
-    def do(self, method: str, path: str, query: dict = None, body: dict = None, raw: bool = False) -> dict:
+    def do(self,
+           method: str,
+           path: str,
+           query: dict = None,
+           body: dict = None,
+           raw: bool = False,
+           files=None,
+           data=None) -> dict:
         headers = {'Accept': 'application/json', 'User-Agent': self._user_agent_base}
         response = self._session.request(method,
                                          f"{self._cfg.host}{path}",
                                          params=query,
                                          json=body,
-                                         headers=headers)
+                                         headers=headers,
+                                         files=files,
+                                         data=data,
+                                         stream=True if raw else False)
         try:
-            if not raw:
-                self._record_request_log(response)
+            self._record_request_log(response, raw=raw or data is not None or files is not None)
             if not response.ok:
                 # TODO: experiment with traceback pruning for better readability
                 # See https://stackoverflow.com/a/58821552/277035
                 payload = response.json()
                 raise self._make_nicer_error(status_code=response.status_code, **payload) from None
+            if raw:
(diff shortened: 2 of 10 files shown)


databricks/sdk/mixins/workspace.py
```diff
@@ -0,0 +1,40 @@
...
+    return f'{x.__module__}.{x.__name__}'
+
+
+class WorkspaceExt(workspace.WorkspaceAPI):
+
+    def upload(self,
+               path: str,
+               content: typing.BinaryIO,
+               *,
+               format: ExportFormat = ExportFormat.AUTO,
```
jerryjam-db, line 19: This technically has different semantics than the Workspace API, which when the format is not specified then it will default to `format: AUTO`. 

Not opposed to this, but just want to make sure that this is a conscious choice.

databricks/sdk/mixins/workspace.py
jerryjam-db, line 1: Do we include some information in the User Agent to indicate that these requests are coming from the Databricks SDK? This will help for tracking on our side.

jerryjam-db (approved):

databricks/sdk/core.py
```diff
@@ -882,9 +894,11 @@ def _record_request_log(self, response: requests.Response):
             for k, v in request.headers.items():
                 sb.append(f'> * {k}: {self._only_n_bytes(v, self._debug_truncate_bytes)}')
         if request.body:
-            sb.append(self._redacted_dump("> ", request.body))
+            sb.append("> [raw stream]" if raw else self._redacted_dump("> ", request.body))
         sb.append(f'< {response.status_code} {response.reason}')
-        if response.content:
+        if raw and 'Content-Type' in response.headers:
```
pietern, line 899: Why check for `Content-Type` header here?

databricks/sdk/mixins/files.py
```diff
@@ -385,3 +393,18 @@ def move_(self, src: str, dst: str, *, recursive=False, overwrite=False):
...
         source.delete(recursive=recursive)
+
+
+class FilesMixin:
+
+    def __init__(self, api_client: ApiClient):
+        self._api = api_client
+
+    def upload(self, path: str, src: BinaryIO):
+        self._api.do('PUT', f'/api/2.0/fs/files{path}', data=src) # files for the workspace upload
```
pietern, line 404: Stale comment?

databricks/sdk/core.py
```diff
@@ -825,20 +825,32 @@ def _authenticate(self, r: requests.PreparedRequest) -> requests.PreparedRequest
...
         try:
-            self._record_request_log(response)
+            self._record_request_log(response, raw=raw or data is not None or files is not None)
             if not response.ok:
                 # TODO: experiment with traceback pruning for better readability
                 # See https://stackoverflow.com/a/58821552/277035
                 payload = response.json()
                 raise self._make_nicer_error(status_code=response.status_code, **payload) from None
+            if raw:
+                return response.raw
```
pietern, line 853: If the caller ever needs access to the content type here, we'd need to return the response object rather than the body stream. Alternatively, we could return both here.

databricks/sdk/mixins/files.py
```diff
@@ -385,3 +393,18 @@ def move_(self, src: str, dst: str, *, recursive=False, overwrite=False):
...
+
+    def download(self, path: str) -> BinaryIO:
+        return self._api.do('GET', f'/api/2.0/fs/files{path}', raw=True)
```
mgyucht, line 407: Is path always absolute?

databricks/sdk/mixins/workspace.py
```diff
@@ -0,0 +1,40 @@
...
+        if not isinstance(format, ExportFormat):
+            raise ValueError(
+                f'format is expected to be {_fqcn(ExportFormat)}, but got {_fqcn(format.__class__)}')
+        if language and not isinstance(language, Language):
+            raise ValueError(
+                f'language is expected to be {_fqcn(Language)}, but got {_fqcn(language.__class__)}')
+        data = {'path': path, 'format': format.value}
+        if language:
+            data['language'] = language.value
+        return self._api.do('POST', '/api/2.0/workspace/import', files={'content': content}, data=data)
```
pietern, line 30: The `data` kwarg in requests appears to be for form data, not JSON body.

What is the resulting request content type here?

pietern: Separate question: does this same API work for uploading regular files? I notice we use the `/workspace-files//import-file` API elsewhere to import regular files.

databricks/sdk/mixins/files.py
mgyucht, line 404: These endpoints don't appear yet in the open API spec, is that intended?

.codegen/__init__.py.tmpl
```diff
@@ -34,6 +35,7 @@ class WorkspaceClient:
         self.config = config
         self.dbutils = dbutils.RemoteDbUtils(self.config)
         self.api_client = client.ApiClient(self.config)
+        self.files = FilesMixin(self.api_client)
```
mgyucht, line 38: Does this need to be added here explicitly because there is no Files tag in the OpenAPI spec?

mgyucht (commented):
Few questions. Nice to unify these things together.

databricks/sdk/core.py
nfx: we don't have such requirements at the moment. returning two results from a function is a breaking change. this will return `io.BinaryIO` or `dict[str,any]`

databricks/sdk/mixins/workspace.py
nfx: image

databricks/sdk/mixins/files.py
nfx: @bogdanghita-db didn't create one yet

.codegen/__init__.py.tmpl
nfx: yes, temporarily

databricks/sdk/core.py
nfx: Raw streams with `Transfer-Encoding: chunked` do not have `Content-Type` header

databricks/sdk/mixins/files.py
nfx: yep

databricks/sdk/mixins/workspace.py
nfx: @jerryjam-db i've added path extension checking in the SDK, so that by default we can have this simple code to import the notebook without having to pass in the `language` parameter.

```python
    py = f'/Users/{w.current_user.me().user_name}/notebook-{random(12)}.py'

    w.workspace.upload(py, io.BytesIO(b'print(1)'))
    with w.workspace.download(py) as f:
        content = f.read()
        assert content == b'# Databricks notebook source\nprint(1)'
```

files are done via:

```python
    py_file = f'/Users/{w.current_user.me().user_name}/file-{random(12)}.py'

    w.workspace.upload(py_file, io.BytesIO(b'print(1)'), format=ExportFormat.AUTO)
    with w.workspace.download(py_file) as f:
        content = f.read()
        assert content == b'print(1)'

    w.workspace.delete(py_file)
```

PR #152, opened 2023-06-07
Improved notebook-native authentication

### LINKED ISSUE #153: Use the sdk from within databricks?
Hello,

I am looking for the simplest to automate submitting a large amount of jobs. I see that databricks_cli https://docs.databricks.com/dev-tools/cli/index.html does not seem to be in active development, so I am looking at this project as a solution for automation.

I would like to be able to run this sdk from within my databricks workspace, but when I try the example https://docs.databricks.com/dev-tools/sdk-python.html#get-started-with-the-databricks-sdk-for-python, I get the error message `default auth: cannot configure default credentials`

I assume that is because databricks-sdk-py is unaware that I am running from within my databricks workspace. My question is, is there a way to make the module recognize that and assume my credentials?

nfx:
Ported from [MLflow codebase](https://github.com/mlflow/mlflow/blame/6bd97bde24d78bcfbf6d50c1dd0f4fac2ed6987b/mlflow/utils/databricks_utils.py#L33):
image

Resolves #153.

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -91,12 +91,24 @@ def pat_auth(cfg: 'Config') -> HeaderFactory:
 @credentials_provider('runtime', [])
 def runtime_native_auth(cfg: 'Config') -> Optional[HeaderFactory]:
     from databricks.sdk.runtime import init_runtime_native_auth
-    if init_runtime_native_auth is None:
-        return None
-    else:
+    if init_runtime_native_auth is not None:
         host, inner = init_runtime_native_auth()
         cfg.host = host
         return inner
+    try:
+        from dbruntime.databricks_repl_context import get_context
+        ctx = get_context()
+        if ctx is None:
+            logger.debug('Empty REPL context returned, skipping runtime auth')
+            return None
+        cfg.host = f'https://{ctx.browserHostName}'
+
+        def inner() -> Dict[str, str]:
+            return {'Authorization': f'Bearer {ctx.apiToken}'}
+
+        return inner
+    except ImportError:
+        return None
 
 
 @credentials_provider('oauth-m2m', ['is_aws', 'host', 'client_id', 'client_secret'])
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -104,6 +104,7 @@ def runtime_native_auth(cfg: 'Config') -> Optional[HeaderFactory]:
         cfg.host = f'https://{ctx.browserHostName}'
 
         def inner() -> Dict[str, str]:
+            ctx = get_context()
             return {'Authorization': f'Bearer {ctx.apiToken}'}
 
         return inner
--- a/databricks/sdk/runtime/__init__.py
+++ b/databricks/sdk/runtime/__init__.py
@@ -16,6 +16,18 @@ try:
 except ImportError:
     init_runtime_native_auth = None
 
+if init_runtime_native_auth is None:
+    try:
+        from dbruntime.databricks_repl_context import get_context
+        ctx = get_context()
+        if ctx is not None:
+
+            def init_runtime_native_auth():
+                inner = lambda: {'Authorization': f'Bearer {ctx.apiToken}'}
+                return f'https://{ctx.browserHostName}', inner
+    except ImportError:
+        init_runtime_native_auth = None
+
 globals()["init_runtime_native_auth"] = init_runtime_native_auth
 
 try:
--- a/databricks/sdk/runtime/__init__.py
+++ b/databricks/sdk/runtime/__init__.py
@@ -16,18 +16,6 @@ try:
 except ImportError:
     init_runtime_native_auth = None
 
-if init_runtime_native_auth is None:
-    try:
-        from dbruntime.databricks_repl_context import get_context
-        ctx = get_context()
-        if ctx is not None:
-
-            def init_runtime_native_auth():
-                inner = lambda: {'Authorization': f'Bearer {ctx.apiToken}'}
-                return f'https://{ctx.browserHostName}', inner
-    except ImportError:
-        init_runtime_native_auth = None
-
 globals()["init_runtime_native_auth"] = init_runtime_native_auth
 
 try:


databricks/sdk/runtime/__init__.py
```diff
@@ -16,6 +16,18 @@
 except ImportError:
     init_runtime_native_auth = None
 
+if init_runtime_native_auth is None:
+    try:
+        from dbruntime.databricks_repl_context import get_context
+        ctx = get_context()
```
gengliangwang, line 22: Shall we try reading the env variable `DATABRICKS_TOKEN` as well. We put it in multiple documentations.
https://docs.databricks.com/dev-tools/python-sql-connector.html
https://python.langchain.com/en/latest/integrations/databricks.html

databricks/sdk/runtime/__init__.py
```diff
@@ -16,6 +16,18 @@
...
+        if ctx is not None:
+
+            def init_runtime_native_auth():
+                inner = lambda: {'Authorization': f'Bearer {ctx.apiToken}'}
+                return f'https://{ctx.browserHostName}', inner
+    except ImportError:
+        init_runtime_native_auth = None
+
 globals()["init_runtime_native_auth"] = init_runtime_native_auth
```
MrBago, line 31: what is this? This is the same as `init_runtime_native_auth = init_runtime_native_auth`

databricks/sdk/runtime/__init__.py
```diff
@@ -16,6 +16,18 @@
 except ImportError:
```
MrBago, line 16: We should not append init_runtime_native_auth to dbruntime_objects

databricks/sdk/runtime/__init__.py
```diff
@@ -16,6 +16,18 @@
...
     init_runtime_native_auth = None
 
+if init_runtime_native_auth is None:
+    try:
+        from dbruntime.databricks_repl_context import get_context
+        ctx = get_context()
+        if ctx is not None:
```
MrBago, line 23: This is a static check on ctx done on import, but `get_context` will not return a Context until a command has been run. This probably works because databricks.sdk.rutnime is probably not imported until a user interacts with the sdk, but this is pretty brittle and will break if we internally import this file on startup in the future.

Can we make this check either dynamic or at if we want to check once and then give up, could we be more explcit about when the check is done.

MrBago (changes requested):
Can we move all the `init_runtime_native_auth` and `databricks_repl_context` logic to a new file, there is no reason any of this needs to be `databricks.sdk.runtime`. The auth stuff and the runtime stuff are distinct and we explicitly do not want them to interact. All this code should work the same in another file.

databricks/sdk/runtime/__init__.py
nfx: it's handled generically in the earlier stage of config resolution: https://github.com/databricks/databricks-sdk-py/blob/main/databricks/sdk/core.py#L704-L717

databricks/sdk/runtime/__init__.py
nfx: this is the part that @xiaochen-db is driving, outside of scope of this PR :)

databricks/sdk/runtime/__init__.py
nfx: this is the part that @xiaochen-db is driving, outside of scope of this PR :)

databricks/sdk/runtime/__init__.py
nfx: addressed in https://github.com/databricks/databricks-sdk-py/pull/152/commits/f37186556cd70411c3188f43a42fc797553af811 - we need a hostname during the SDK initialization stage. technically, if we can start every runtime with `DATABRICKS_HOST` environment variable with the f'https://{ctx.browserHostName}', we can simplify the `init_runtime_native_auth` return type. 

please remember - `init_runtime_native_auth` returns the workspace hostname and function that returns a dictionary of headers. this is to support future improvements to REPL authentication, where we could leverage token refreshes

gengliangwang (approved):

databricks/sdk/core.py
```diff
@@ -91,12 +91,24 @@ def inner() -> Dict[str, str]:
...
     from databricks.sdk.runtime import init_runtime_native_auth
-    if init_runtime_native_auth is None:
-        return None
-    else:
+    if init_runtime_native_auth is not None:
         host, inner = init_runtime_native_auth()
         cfg.host = host
         return inner
+    try:
+        from dbruntime.databricks_repl_context import get_context
```
dby-tmwctw, line 99: Hi @nfx may I know whether you verify in notebook environment (e.g. with DBR 12.2) that it works? I vaguely remember last time I want to import from `dbruntime` in `core.py` it does not work. Ignore me if I'm wrong.

databricks/sdk/core.py
```diff
@@ -91,12 +91,24 @@ def inner() -> Dict[str, str]:
...
+        ctx = get_context()
+        if ctx is None:
+            logger.debug('Empty REPL context returned, skipping runtime auth')
+            return None
+        cfg.host = f'https://{ctx.browserHostName}'
+
+        def inner() -> Dict[str, str]:
+            return {'Authorization': f'Bearer {ctx.apiToken}'}
```
xiaochen-db, line 107: Don't we lock in the `apiToken`? I think we should call `get_context()` inside `inner` to swap the context for different users.

nfx: @xiaochen-db also running `get_context()` within the `inner`? thanks for pointing this out! will add

databricks/sdk/core.py
nfx: I've tested it on couple of DBRs

databricks/sdk/core.py
nfx: addressed in https://github.com/databricks/databricks-sdk-py/pull/152/commits/b95af594b9d0ee519e03ebc584c1aeac1f1f51f3

databricks/sdk/core.py
```diff
@@ -91,12 +91,25 @@ def inner() -> Dict[str, str]:
...
-    if init_runtime_native_auth is None:
-        return None
-    else:
+    if init_runtime_native_auth is not None:
         host, inner = init_runtime_native_auth()
         cfg.host = host
         return inner
+    try:
+        from dbruntime.databricks_repl_context import get_context
+        ctx = get_context()
```
xiaochen-db, line 100: Just want to double check when is this called? Inside a command or during REPL initialization?

lennartkats-db (approved):

MrBago (approved):

databricks/sdk/core.py
nfx: both

PR #155, opened 2023-06-08
Add "wait" to run_now_jobs_api_full_integration

lennartkats-db:
## Changes

@nfx  Looking at this code, I think it should use the "and wait" method?


nfx (changes requested):
no, the code is correct. `run_now()` starts the run and `.result()` sleeps/waits. this example is re-generated from Go SDK nightly integration tests, so changing this particular one is not advised, as changes will be lost

PR #161, opened 2023-06-09
(bugfix) Catalog GrantAPI: securable_type enum value

clementboscoa:
## Changes
Change the render value of securable_type attribute to have the real Enum value and not the __repr__ value.
This will render the API URI call as the following : PATCH /api/2.1/unity-catalog/permissions/CATALOG/<catalog_id>
instead of /api/2.1/unity-catalog/permissions/SecurableType.CATALOG/<catalog_id>

## Tests
Tested locally

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/databricks/sdk/service/catalog.py
+++ b/databricks/sdk/service/catalog.py
@@ -3498,7 +3498,7 @@ class GrantsAPI:
 
         json = self._api.do(
             'GET',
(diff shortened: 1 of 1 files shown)


nfx:
correctly implemented in https://github.com/databricks/databricks-sdk-py/pull/179

PR #168, opened 2023-06-14
Correctly parse `libraries` in `service.jobs`

judahrand:
## Changes

Pretty self explanatory. Currently, this is correct in most places but is wrong in the `jobs` service. I suspect there is either a bug in the `service.py.tmpl` or a mistake in the OpenAPI spec.

I'd debug it myself but since the OpenAPI spec is available anywhere in raw form that's a bit tricky. Could you make in available, please?

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/databricks/sdk/service/jobs.py
+++ b/databricks/sdk/service/jobs.py
@@ -189,7 +189,7 @@ class ClusterSpec:
     @classmethod
     def from_dict(cls, d: Dict[str, any]) -> 'ClusterSpec':
         return cls(existing_cluster_id=d.get('existing_cluster_id', None),
-                   libraries=d.get('libraries', None),
+                   libraries=_repeated(d, 'libraries', Library),
                    new_cluster=_from_dict(d, 'new_cluster', BaseClusterInfo))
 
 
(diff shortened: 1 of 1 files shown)


nfx:
this file is auto-generated and will silently be rewritten in other releases. we'll have to fix bug on generator template side

judahrand:
> I'd debug it myself but since the OpenAPI spec is available anywhere in raw form that's a bit tricky. Could you make in available, please?

I'd debug it myself but since the OpenAPI spec is available anywhere in raw form that's a bit tricky. Could you make in available, please?

nfx:
@judahrand it's proprietary for now :)

judahrand:
> @judahrand it's proprietary for now :)

Well sort of... it could mostly be reverse engineered from this SDK I'd guess. Also seems like an unhelpful thing to make proprietary given that it defines how to interact with the Databricks service. Surely Databricks wants as many services as possible to integrate with Databricks?

You used to publish your JobsAPI 2.1 YAML... but that seems to have been yanked.

nfx:
@judahrand we first want a few good official SDKs that work the same, configure the same, and feel the same and are available for every major language :) thanks for your bug reports!

judahrand:
I suspect that the issue might actually be in the OpenAPI definitions for the Jobs API since it only seems to happen for the Jobs service and not other services.

nfx:
Fixing with more entities in https://github.com/databricks/databricks-sdk-py/pull/178

PR #170, opened 2023-06-14
Mark non-required fields as `Optional`

judahrand:
## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/.codegen/service.py.tmpl
+++ b/.codegen/service.py.tmpl
@@ -3,7 +3,7 @@
 from dataclasses import dataclass
 from datetime import timedelta
 from enum import Enum
-from typing import Dict, List, Any, Iterator, Type, Callable
+from typing import Dict, List, Any, Iterator, Type, Callable, Optional
 import time
 import random
 import logging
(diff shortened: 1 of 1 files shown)


nfx:
@judahrand files that start with `# Code generated from OpenAPI specs by Databricks SDK Generator. DO NOT EDIT.` will be silently overridden. please create github issues for those instead.

https://github.com/databricks/databricks-sdk-py/blob/main/databricks/sdk/service/jobs.py#L1

nfx:
@judahrand you can also try to add small modifications to https://github.com/databricks/databricks-sdk-py/blob/main/.codegen/service.py.tmpl#L1-L2, but the proprietary dataset is not available to the public and there's no way to test it if you're not working for Databricks.

judahrand:
> @judahrand files that start with `# Code generated from OpenAPI specs by Databricks SDK Generator. DO NOT EDIT.` will be silently overridden. please create github issues for those instead.
> 
> https://github.com/databricks/databricks-sdk-py/blob/main/databricks/sdk/service/jobs.py#L1

This one is the template isn't it? So this particular file isn't autogenerated if my understanding is correct?

nfx (approved):

PR #172, opened 2023-06-14
Create a deepcopy of config in api client

kartikgupta-db:
## Changes
* We do not want the config to change after API client has been initialised. One API client should be linked to 1 set of Config variables ONLY.

- [x] `make test` run locally
- [x] `make fmt` applied


databricks/sdk/__init__.py
```diff
@@ -58,6 +60,11 @@
 
 
 class WorkspaceClient:
+    _config: client.Config
+
+    @property
+    def config(self):
+        return deepcopy(self._config)
```
nfx, line 67: i'm not sure it's a good thing. `w.config.authenticate()` should reuse the same refresh token

nfx (changes requested):

databricks/sdk/__init__.py
kartikgupta-db: Riggghttt. Completely forgot about this. We can atleast restrict the initial config form being passed around. If they directly reuse config from client, or edit it, atleast they know why stuff is changing.

main.py
```diff
@@ -0,0 +1,5 @@
+from databricks.connect import DatabricksSession
+from databricks.sdk.core import Config
+
+config = Config(host='https://x', token='x', product="xapp", product_version="0.0.0")
```
nfx, line 4: please remove this file

nfx (changes requested):
i'm still not sure that we should do it this way. if we want to prevent config from being modified, we can add these in Config:

```python
    self._init_auth()
    self._initialized = True
```

and these in `ConfigAttribute`:

```python
    def __set__(self, cfg: 'Config', value: any):
        if cfg._initialized: raise ValueError(...)
        cfg._inner[self.name] = self.transform(value)
```

and test in in DBR for dbr-native auth

nfx (approved):

PR #173, opened 2023-06-14
Add basic rate limiting

nfx:
Adds basic implementation of client-side rate limiting with no dependencies outside of standard library.

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -1,5 +1,6 @@
 import abc
 import base64
+import collections
 import configparser
 import functools
 import json
@@ -10,6 +11,8 @@ import platform
 import re
 import subprocess
 import sys
+import threading
+import time
 import urllib.parse
 from datetime import datetime
 from json import JSONDecodeError
@@ -849,6 +852,30 @@ class DatabricksError(IOError):
         self.kwargs = kwargs
 
 
+class RateLimiter:
+    def __init__(self, max_requests: int):
+        if max_requests is None:
+            max_requests = 30
+        self._max_requests = int(max_requests)
+        self._lock = threading.Lock()
+        self._requests = collections.deque()
+
+    def _window_size(self):
+        return self._requests[-1] - self._requests[0]
+
+    def throttle(self):
+        with self._lock:
+            burst_period_seconds = 1
+            if len(self._requests) >= self._max_requests:
+                sleep = self._window_size()
+                logger.debug(f'Request throttled. Sleeping for {sleep}s')
+                time.sleep(sleep)
+            self._requests.append(time.time())
+            while self._window_size() >= burst_period_seconds:
+                self._requests.popleft()
+            return self
+
+
 class ApiClient:
     _cfg: Config
 
@@ -873,6 +900,7 @@ class ApiClient:
         self._session = requests.Session()
         self._session.auth = self._authenticate
         self._session.mount("https://", HTTPAdapter(max_retries=retry_strategy))
+        self._rate_limiter = RateLimiter(cfg.rate_limit)
 
     @property
     def account_id(self) -> str:
@@ -904,6 +932,7 @@ class ApiClient:
            raw: bool = False,
            files=None,
            data=None) -> dict:
+        self._rate_limiter.throttle()
         headers = {'Accept': 'application/json', 'User-Agent': self._user_agent_base}
         response = self._session.request(method,
                                          f"{self._cfg.host}{path}",
--- a/tests/integration/test_clusters.py
+++ b/tests/integration/test_clusters.py
@@ -6,6 +6,15 @@ import pytest
 from databricks.sdk.core import DatabricksError
 
 
+def test_noise(w):
+    from concurrent.futures import ThreadPoolExecutor, wait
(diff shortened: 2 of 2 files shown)


databricks/sdk/core.py
```diff
@@ -849,6 +852,30 @@ def __init__(self,
...
+            if len(self._requests) >= self._max_requests:
+                sleep = self._window_size()
+                logger.debug(f'Request throttled. Sleeping for {sleep}s')
+                time.sleep(sleep)
+            self._requests.append(time.time())
+            while self._window_size() >= burst_period_seconds:
+                self._requests.popleft()
+            return self
+
+
```
pietern, line 878: Please add a reference to where the algorithm comes from, where else it is used, etc.

I'd also like to see a couple narrowed tests than the 100 API calls test below. You could make the burst time a constructor kwarg to make those tests run fast.

nfx: will do, it's more to get a gauge if we want to do it this way. 

This feature is aimed for v0.1.11

ericfeunekes: FYI, another reference implementation that I came up with a while back. I haven't benchmarked it, but it might be a bit more performant since it doesn't use a lock.

```python 
import time
from asyncio import Queue, sleep
from functools import wraps
from typing import Any, Callable, Coroutine


def limiter(
    num_calls: int, time_limit: int
) -> Callable[..., Callable[..., Coroutine[Any, Any, Any]]]:
    """Decorator that limits the rate at which the decorated function can be called.

    Parameters:
        num_calls (int): The maximum number of calls to the decorated function allowed
        within a time window.
        time_limit (int): The time limit in seconds of each time window.

    Returns:
        Callable: A function that decorates another function to limit the rate at which
        it can be called.

    Example usage:
        ```
        @limiter(num_calls=10, time_limit=60)
        async def my_function():
            # Do some stuff here
        ```
        limits `my_function to 10 calls per 60 seconds
    """
    sleep_queue = create_sleep_queue(num_calls, time_limit)

    def outer(
        func: Callable[..., Coroutine[Any, Any, Any]]
    ) -> Callable[..., Coroutine[Any, Any, Any]]:
        @wraps(func)
        async def inner(*args: Any, **kwargs: Any) -> Any:
            nonlocal sleep_queue

            sleep_time = await sleep_queue.get()
            try:
                if time.time_ns() < sleep_time:
                    await sleep((sleep_time - time.time_ns()) / 1_000_000_000)
                result = await func(*args, **kwargs)
            finally:
                await sleep_queue.put(time.time_ns() + time_limit * 1_000_000_000)

            return result

        return inner

    return outer


def create_sleep_queue(num_calls: int, time_limit: int) -> Queue[int]:
    """Create a queue of num_calls sleep times.

    Parameters:
        num_calls (int): The maximum number of sleep times to create.
        time_limit (int): The time limit in seconds for the sleep.

    Returns:
        Queue[int]: A queue of sleep times.
    """
    sleep_queue: Queue[int] = Queue(maxsize=num_calls)
    past_time = time.time_ns() - 99 * 365 * 24 * 60 * 60 * 1_000_000_000

    for _ in range(num_calls):
        sleep_queue.put_nowait(past_time)

    return sleep_queu
```

nfx: @ericfeunekes we don't support `asyncio` yet, as the entire auth stack has to be written for it from scratch. 

If community is interested in `asyncio`-based Python SDK for Databricks, then we'd be happy to accept contributions and guide in the process.

databricks/sdk/core.py
```diff
@@ -849,6 +852,30 @@ def __init__(self,
...
+        with self._lock:
+            burst_period_seconds = 1
+            if len(self._requests) >= self._max_requests:
+                sleep = self._window_size()
+                logger.debug(f'Request throttled. Sleeping for {sleep}s')
+                time.sleep(sleep)
+            self._requests.append(time.time())
+            while self._window_size() >= burst_period_seconds:
+                self._requests.popleft()
+            return self
```
mgyucht, line 876: Maybe I'm misunderstanding the algorithm, but what makes sense to me is:
```
start_time = time.time()
while self._requests[0] < start_time - burst_period_seconds
  self._requests.popleft()
if len(self._requests) > self._max_requests:
  sleep_time = burst_time_seconds - self._requests[0]
  time.sleep(sleep_time)
self._requests.append(time.time())
```
Suppose that 30 requests are made within a second. In the current implementation, the next request will always sleep for burst_period_seconds, even if made more than one second later. This version would clear the request queue before checking if sleeping is needed.

mgyucht (commented):
I think there is a slight improvement we can make to prevent unnecessary sleeping.

databricks/sdk/core.py
ericfeunekes: This is just a reference. You don't need asyncio, you can just use a different queue from whatever standard library you want and remove the  `async def`. 

It was more to show a different approach that allows concurrency if you want it, without locking for the duration of the sleep. Even in a non-concurrent context this allows bursting to the max speed that the network allows. So you can make it non-concurrent now and easily migrate to threaded or async code in the future by changing the queue type.

PR #210, opened 2023-06-29
Make workspace client also return runtime dbutils when in dbr

kartikgupta-db:
## Changes
* `workspace_client.dbutils` always returned oss implementation of dbutils. We want it to also use dbr implementation when in dbr.

## Tests
* [x] Manually Test in dbr
* [x] Test locally 

- [x] `make test` run locally
- [x] `make fmt` applied


mgyucht (approved):
LGTM! I suppose this is as good a test as any for whether we are in DBR or not.

mgyucht:
Suggestion: let's not include internal ticket IDs in PR descriptions in public repos.

.codegen/__init__.py.tmpl
```diff
@@ -17,6 +17,22 @@ from databricks.sdk.service.{{.Package.Name}} import {{.PascalName}}API{{end}}
   {{- getOrDefault $mixins $genApi $genApi -}}
 {{- end -}}
 
+def _make_dbutils(config: client.Config):
```
nfx, line 20: i'm not sure it's the intended behavior, as we want to migrate users away from dbutils to SDK. this PR makes people assume that dbutils will stay there forever

nfx (changes requested):

.codegen/__init__.py.tmpl
kartikgupta-db: Agreed. But there is a currently a huge functionality gap between the sdk implementation and dbutils on runtime. 
1. Having the 2 entrypoints (`from databricks.sdk.runtime import *` and workspace_client.dbutils) provide different implementations can be confusing for the user. 
2. Also, it doesn't make sense to use the command-exec api (for proxyutils) when we are already in the runtime. And there is functionality (such as widgets) which will necessarily work differently in runtime and sdk.

If we are ok with putting the responsibility on the user to understand what dbutils they are using (1.), we can address 2 by selectively proxying to the dbr dbutils when in dbr (instead of using command exec or different implementations for widgets).

nfx: @xiaochen-db @falaki - what do you guys think?

xiaochen-db: Sounds like we are reversing from the former direction where REPL should eventually adopt the SDK dbutils despite current function gaps to maintaining two versions of dbutils: the SDK one and the DBR one, and SDK one will fall back on DBR if it's used in a REPL. 

I am okay with that given that the function gaps seem unresolvable in the near term. My another question is that why do you only update the dbutils from `WorkspaceClient`? Is there other way for a user to create `dbutils.RemoteDbUtils(config)`?

kartikgupta-db: We expose dbutils through 2 paths
1. `from databricks.sdk.runtime import *`
2. workspaceClient.dbutils

The first one already falls back to the dbr version of dbutils. WorkspaceClient does not.

Good point about `dbutils.RemoteDbUtils(config)`. It is possible. It currently looks like this
```
from databricks.sdk import dbutils
dbutils.RemoteDbUtils(config).fs.ls(".")
```

We should rename this to `_dbutils` to make it clear that this is an internal implementation, and users are not expected to initialise it. Wdyt @nfx ?

xiaochen-db (approved):

.codegen/__init__.py.tmpl
kartikgupta-db: @judahrand this will already work with the changes you made. This will return a union of types, which will include the full typing information you added to sdk.runtime

judahrand: I meant type hinting in the code 🤷‍♂️ No reason not to properly type hint all Python code these days.

kartikgupta-db: @judahrand it is difficult to type this function, because we cannot know what type of dbutils to use (the dbr or the OSS one)  until we actually try importing dbutils from dbr. So this function can't be typed until this function runs. 

We do not want to mess with dbr typing.

PR #213, opened 2023-07-03
Strip quotes from env vars

kartikgupta-db:
## Changes
* SDK does not strip quotes when loading from environment variables. This is based on the assumption that os.environ already handles quote stripping. Something like `profile="something"` works. 
* But, there are cases where quotes might be escaped (not on purpose). Eg, when copying environment variables to pycharm, if the value has quotes, the quotes are escaped. 
image
is parsed as 
image
**Note: '"10"'**

This is not handled properly by sdk. The quotes are note stripped. This PR adds stripping for these. 

## Tests
- [ ] `make test` run locally
- [ ] `make fmt` applied


mgyucht (changes requested):
As discussed, we will quote variables used in logs/exceptions to make it clear whether parameters are quoted or not.

renaudhartert-db:

PR #219, opened 2023-07-07
Rearrange imports to improve local editor experience

judahrand:
## Changes

The type hinting here means that VSCode does not give useful syntax highlights / code completions. 

This is the current experience on `main`:
image
image

With these changes this becomes:
image
image
 

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


nfx:
@MrBago @kartikgupta-db could you take a look if it doesn't break anything

kartikgupta-db (approved):
LGTM for the OSS part.

nfx:
@MrBago @xiaochen-db need review/approval from DBR part

databricks/sdk/runtime/__init__.py
```diff
@@ -31,15 +33,22 @@
...
+        from . import stub
         from .stub import *
+        dbutils_type = Type[stub.dbutils]
     except (ImportError, NameError):
         from databricks.sdk.dbutils import RemoteDbUtils
 
         # this assumes that all environment variables are set
         dbutils = RemoteDbUtils()
+        dbutils_type = RemoteDbUtils
+
```
xiaochen-db, line 51: Can you enlighten me if this following branch 
```
    try:
        from . import stub
        from .stub import *
        dbutils_type = Type[stub.dbutils]
```

doesn't throw any exception, then what is `dbutils`? From stub.py it's a class, then why are you casting `dbutils` (class) to `dbutils_type` (Type)?

judahrand: We've defined `dbutils_type` in two places and so it ends up being a Union of the class from the stub and the `ProxyCall` class. We then cast to this Union.

MrBago: Do we really want the union type? It seems like the "RemoteDbutils" type doesn't have useful info for the user, can we just suppress it?

MrBago (approved):

PR #226, opened 2023-07-10
Fix job delete call in run_now_jobs_api_full_integration.py

slakshmangit:
jobs. delete method requires job_id as param.

## Changes
Modified delete job call by replacing the param with job_id instead of delete.

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied


mgyucht:
@slakshmangit thank you for contributing to the Python SDK! This example is autogenerated from integration tests in the Go SDK, so this change would be overwritten the next time we release the Python SDK. Because of this, I'm going to close your PR. If you're interested, you can see the PR that fixes the underlying issue here: https://github.com/databricks/databricks-sdk-go/pull/560.

PR #229, opened 2023-07-12
Regenerate Python SDK using recent OpenAPI Specification

mgyucht:
## Changes
Spec commit sha: 17a3f7fe6 (7 July 2023)

Breaking Changes:
* Use CONSTANT_CASE for Enum constants. Many enums already use constant case in their definition, but some constants (like the SCIM Patch schema name) includes symbols `:` and numbers, so the SDK cannot use the enum value as the name.
* Replace Query type with AlertQuery in sql.Alert class.
* Removal of User.is_db_admin and User.profile_image_url.

Changes:
* Introduce CleanRooms API
* Introduce TablesAPI.update()
* Introduce Group.meta property
* Fix SCIM Patch implementation
* Introduce BaseRun.job_parameters and BaseRun.trigger_info
* Introduce CreateJob.parameters
* Fix spelling in file arrival trigger configuration
* Introduce GitSource.job_source
* Introduce RepairRun.rerun_dependent_tasks
* Introduce Resolved*Values classes, RunIf, and RunJobTask
* Introduce TaskNotificationSettings

Later follow-up:
* Names should split on Pascal-case word boundaries (see CloudProviderNodeStatus). This is an OpenAPI code gen change that needs to be made.


databricks/sdk/service/compute.py
```diff
@@ -209,8 +209,8 @@ def from_dict(cls, d: Dict[str, any]) -> 'CloudProviderNodeInfo':
 
 class CloudProviderNodeStatus(Enum):
 
-    NotAvailableInRegion = 'NotAvailableInRegion'
-    NotEnabledOnSubscription = 'NotEnabledOnSubscription'
+    NOTAVAILABLEINREGION = 'NotAvailableInRegion'
+    NOTENABLEDONSUBSCRIPTION = 'NotEnabledOnSubscription'
```
pietern, line 213: This is a UX regression, IMO. These names should be snake cased.

databricks/sdk/service/compute.py
```diff
@@ -3057,8 +3061,8 @@ class TerminationReasonCode(Enum):
     INVALID_SPARK_IMAGE = 'INVALID_SPARK_IMAGE'
     IP_EXHAUSTION_FAILURE = 'IP_EXHAUSTION_FAILURE'
     JOB_FINISHED = 'JOB_FINISHED'
-    K8S_AUTOSCALING_FAILURE = 'K8S_AUTOSCALING_FAILURE'
-    K8S_DBR_CLUSTER_LAUNCH_TIMEOUT = 'K8S_DBR_CLUSTER_LAUNCH_TIMEOUT'
+    KS_AUTOSCALING_FAILURE = 'K8S_AUTOSCALING_FAILURE'
+    KS_DBR_CLUSTER_LAUNCH_TIMEOUT = 'K8S_DBR_CLUSTER_LAUNCH_TIMEOUT'
```
pietern, line 3065: The number disappeared from the constant name.

databricks/sdk/service/jobs.py
```diff
@@ -489,14 +498,14 @@ class GetRunRequest:
...
-    gitLab = 'gitLab'
-    gitLabEnterpriseEdition = 'gitLabEnterpriseEdition'
+    AWSCODECOMMIT = 'awsCodeCommit'
+    AZUREDEVOPSSERVICES = 'azureDevOpsServices'
+    BITBUCKETCLOUD = 'bitbucketCloud'
+    BITBUCKETSERVER = 'bitbucketServer'
+    GITHUB = 'gitHub'
+    GITHUBENTERPRISE = 'gitHubEnterprise'
+    GITLAB = 'gitLab'
+    GITLABENTERPRISEEDITION = 'gitLabEnterpriseEdition'
```
pietern, line 508: Similar to the other one, it would be nice to snake case these.

pietern (approved):

PR #230, opened 2023-07-13
Use .ConstantName defining target enum states for waiters

mgyucht:
## Changes
Uses of enums in generated code need to be updated to use `{{.ConstantName}}` instead of `{{.Content}}`.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/.codegen/service.py.tmpl
+++ b/.codegen/service.py.tmpl
@@ -113,8 +113,8 @@ class {{.Name}}API:{{if .Description}}
     def {{.SnakeName}}(self{{range .Binding}}, {{.PollField.SnakeName}}: {{template "type-nq" .PollField.Entity}}{{end}},
       timeout=timedelta(minutes={{.Timeout}}), callback: Optional[Callable[[{{.Poll.Response.PascalName}}], None]] = None) -> {{.Poll.Response.PascalName}}:
       deadline = time.time() + timeout.total_seconds()
-      target_states = ({{range .Success}}{{.Entity.PascalName}}.{{.Content}}, {{end}}){{if .Failure}}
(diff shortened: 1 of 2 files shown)


pietern (approved):

.codegen/service.py.tmpl
```diff
@@ -113,8 +113,8 @@ class {{.Name}}API:{{if .Description}}
     def {{.SnakeName}}(self{{range .Binding}}, {{.PollField.SnakeName}}: {{template "type-nq" .PollField.Entity}}{{end}},
       timeout=timedelta(minutes={{.Timeout}}), callback: Optional[Callable[[{{.Poll.Response.PascalName}}], None]] = None) -> {{.Poll.Response.PascalName}}:
       deadline = time.time() + timeout.total_seconds()
-      target_states = ({{range .Success}}{{.Entity.PascalName}}.{{.Content}}, {{end}}){{if .Failure}}
-      failure_states = ({{range .Failure}}{{.Entity.PascalName}}.{{.Content}}, {{end}}){{end}}
+      target_states = ({{range .Success}}{{.Entity.PascalName}}.{{.ConstantName}}, {{end}}){{if .Failure}}
```
nfx, line 116: please also change the example template to use correct constant names - otherwise documentation won't be correct.

nfx (changes requested):

.codegen/service.py.tmpl
mgyucht: That was already updated in the last PR

mgyucht:
```
{{- else if eq .Type "enum" -}}
    {{.Package}}.{{.Entity.PascalName}}.{{.ConstantName}}
```
is there currently

nfx (approved):

PR #232, opened 2023-07-14
Fix/add dataclass permission level

Dellonath:
## Changes
I added the @classdata decorator in iam.PemissionLevel(Enum) class. The PemissionLevel class in unrecognized when using this SDK in a python code.

## Tests

I created the test_permission_level_job function in the test_jobs.py module. I instantiate a cluster, a task, and a permission level setting two permissions configurations.

- [ ] `make test` run locally
- [x] `make fmt` applied

### DIFF
--- a/tests/integration/test_jobs.py
+++ b/tests/integration/test_jobs.py
@@ -112,4 +112,4 @@ def test_permission_level_job(w):
         )
     ]
 
-    w.jobs.create(job_clusters=[cluster], tasks=[task1], access_control_list = [access_control_list])
+    w.jobs.create(job_clusters=[cluster], tasks=[task1], access_control_list = access_control_list)
--- a/tests/integration/test_jobs.py
+++ b/tests/integration/test_jobs.py
@@ -87,3 +87,29 @@ def test_create_job(w):
                       python_wheel_task=jobs.PythonWheelTask(entry_point="test", package_name="deepspeed"))
 
     w.jobs.create(job_clusters=[cluster], tasks=[task1])
+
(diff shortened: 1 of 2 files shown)


tests/integration/test_jobs.py
```diff
@@ -87,3 +87,29 @@ def test_create_job(w):
...
+    cluster = jobs.JobCluster(job_cluster_key="cluster1",
+                              new_cluster=compute.ClusterSpec(
+                                  num_workers=2,
+                                  spark_version=w.clusters.select_spark_version(),
+                                  node_type_id=w.clusters.select_node_type(local_disk=True)))
+    
+    access_control_list = [
+        iam.AccessControlRequest(
+            group_name = 'gp_manage',
+            permission_level = iam.PermissionLevel('CAN_MANAGE')
```
mgyucht, line 107: You should be able to do `iam.PermissionLevel.CAN_MANAGE`.

mgyucht (changes requested):
This is an enum, so you should be able to refer to the enum values in the same way you refer to an attribute of an object. Am I missing something?

tests/integration/test_jobs.py
Dellonath: You're completely right!! Thank you for this. This is my first contribution to an open-source tool. Sorry about that :)

mgyucht: No worries, we're always glad to discuss. If you see other issues in our SDKs, please feel free to file new issues (if existing ones don't already cover the problem) or try to make PRs to address any existing issues.

PR #234, opened 2023-07-17
Fix enum deserialization

mgyucht:
## Changes
In #230, enums were changed so that enum field names did not necessarily match the enum value itself. However, the `_enum` helper method used during deserialization of a response containing an enum was not updated to handle this case. This PR corrects this method to check through the values of the `__members__` of an enum, as opposed to the keys.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


nfx (approved):
Lgtm

PR #235, opened 2023-07-17
Fix enum deserialization, take 2

mgyucht:
## Changes
We jumped the gun too quickly on #234. This is the actual change which fixes the integration tests.

## Tests

- [x] The two failing integration tests (test_submitting_jobs and test_proxy_dbfs_mounts) both pass on this PR.


nfx (approved):
can you run `deco env test-all` next time before you make a change in these files? ;)

PR #216, opened 2023-07-06
Improve way of accessing `__annotations__`

MichaelSpece:
## Changes
Use recommended way of accessing `__annotations__` while maintaining backward compatibility.

## Tests
I've added a test that catches the bug.

- [x] `make test` run locally
- [x] `make fmt` applied
- [ ] relevant integration tests applied

PR #239, opened 2023-07-17
Improve access of `__annotations__`

MichaelSpece:
## Changes
This is a continuation of #216, but with signed, verified commits.

## Tests
One test was added (in #216).  The last run in #216 passed all tests.

- [ ] `make test` run locally
- [ ] `make fmt` applied


MichaelSpece:
@nfx GitHub was complaining about unsigned commits so I resynced my fork, which closed #216.

nfx (changes requested):
this looks like too big of a change, can't accept it.

please make sure the PR holds diff only for the relevant changes. most likely the newline characters are changed or something like it.

MichaelSpece:
@nfx Ok; the line endings had indeed been modified.  I have now rebased with auto handling of line endings and the diff is showing a relatively small difference.

tests/test_core.py
```diff
@@ -207,3 +207,19 @@ def test_config_accounts_dod_is_accounts_host(config):
...
+        def __init__(self):
+            super().__init__()
+
+    with pytest.raises(ValueError): # As opposed to `KeyError`.
+        DatabricksConfig()
+
+
+if __name__ == "__main__":
+    import conftest
+    test_config_accounts_dod_is_accounts_host(conftest.config.__pytest_wrapped__.obj())
```
nfx, line 225: these seem irrelevant

nfx (changes requested):

PR #240, opened 2023-07-18
Support older versions of `urllib3` and Databricks Runtime with regards to `DEFAULT_METHOD_WHITELIST` change to `DEFAULT_ALLOWED_METHODS`

mgyucht:
## Changes
This PR changes the SDK to tolerate urllib3 < 1.26.0. In these versions of urllib3, Retry class's `DEFAULT_ALLOWED_METHODS` was called `DEFAULT_METHOD_WHITELIST`, and the parameter in Retry's constructor was changed from `method_whitelist` to `allowed_methods`. With this change, users using older versions of DBR (like the 7.3 LTS) can still use the current version of the SDK, even though the version of urllib3 is old.

This change can be reverted when we no longer need to support urllib3 < 1.26.0 (i.e. after DBR 10.4 LTS, which is EOL on [March 18, 2025](https://docs.databricks.com/release-notes/runtime/releases.html)).

## Test

- [x] Manually tested on DBR 7.3LTS: Old_DBR_test_-_Databricks
- [x] Manually tested on DBR 13.1: 
Old_DBR_test_-_Databricks_2


nfx (approved):
lgtm

pietern (approved):

Issue #99, opened 2023-05-08
Query History list() does not pass filter_by correctly


asikowitz:
Although it's not documented, the query history list [endpoint](https://docs.databricks.com/api-explorer/workspace/queryhistory/list) seems to handle `filter_by` properly only if it is passed in the request body rather than as a query param:

Ran using databricks-sdk v0.1.5, requests v2.28.2:
```
In [2]: from databricks.sdk import WorkspaceClient
In [3]: client = WorkspaceClient(host=..., token=...)
In [4]: from databricks.sdk.service.sql import QueryFilter
In [6]: filter_by = QueryFilter.from_dict(
   ...:             {
   ...:                 "query_start_time_range": {
   ...:                     "start_time_ms": 0,
   ...:                     "end_time_ms": int(time.time() * 1000),
   ...:                 }
   ...:             }
   ...:         )

In [7]: filter_by
Out[7]: QueryFilter(query_start_time_range=TimeRange(end_time_ms=1683569801679, start_time_ms=0), statuses=None, user_ids=None, warehouse_ids=None)

In [9]: next(client.query_history.list(filter_by=filter_by))
---------------------------------------------------------------------------
DatabricksError                           Traceback (most recent call last)
Cell In[9], line 1
----> 1 next(client.query_history.list(filter_by=filter_by))

File ~/.pyenv/versions/3.10.9/lib/python3.10/site-packages/databricks/sdk/service/sql.py:2824, in QueryHistoryAPI.list(self, filter_by, include_metrics, max_results, page_token, **kwargs)
   2821 if page_token: query['page_token'] = request.page_token
   2823 while True:
-> 2824     json = self._api.do('GET', '/api/2.0/sql/history/queries', query=query)
   2825     if 'res' not in json or not json['res']:
   2826         return

File ~/.pyenv/versions/3.10.9/lib/python3.10/site-packages/databricks/sdk/core.py:753, in ApiClient.do(self, method, path, query, body)
    749 if not response.ok:
    750     # TODO: experiment with traceback pruning for better readability
    751     # See https://stackoverflow.com/a/58821552/277035
    752     payload = response.json()
--> 753     raise self._make_nicer_error(status_code=response.status_code, **payload) from None
    754 if not len(response.content):
    755     return {}

DatabricksError: Could not parse request object: Expected 'START_OBJECT' not 'VALUE_STRING'
 at [Source: (ByteArrayInputStream); line: 1, column: 15]
[...]

PR #244, opened 2023-07-20
fix(sql): workaround for filtering in query history api

lukehsiao:
fix(sql): workaround for filtering in query history api

This patch is intended to be a temporary workaround for the
inconsistency of the REST API [[1]]. The REST API states that all inputs
are query parameters, but in reality, they must be in the body.

This also works around the behavior of `filter_by` ignoring the
`page_token` when provided together by dropping the filter_by in
subsequent requests.

[1]: https://docs.databricks.com/api/workspace/queryhistory/list

Tested: Ran the Python SDK locally and confirmed I was able to filter
  and use the other parameters. Example code which filters by the
  query start time.

    from databricks.sdk.service.sql import QueryFilter, QueryInfo, TimeRange
    w = WorkspaceClient(host=redacted, token=redacted)

    filter = QueryFilter(
        query_start_time_range=TimeRange(
            start_time_ms=1689626553081,
            end_time_ms=1689626553083,
        )
    )
    assert len(list(w.query_history.list(filter_by=filter))) == 1
    filter = QueryFilter(
        query_start_time_range=TimeRange(
            start_time_ms=1689626553083,
            end_time_ms=1689626553083,
        )
    )
    assert len(list(w.query_history.list(filter_by=filter))) == 0

Closes: https://github.com/databricks/databricks-sdk-py/issues/99

---

- [x] `make test` run locally
- [x] `make fmt` applied

### DIFF
--- a/databricks/sdk/service/sql.py
+++ b/databricks/sdk/service/sql.py
@@ -3101,21 +3101,25 @@ class QueryHistoryAPI:
                                               max_results=max_results,
                                               page_token=page_token)
 
-        query = {}
-        if filter_by: query['filter_by'] = request.filter_by.as_dict()
-        if include_metrics: query['include_metrics'] = request.include_metrics
-        if max_results: query['max_results'] = request.max_results
-        if page_token: query['page_token'] = request.page_token
+        body = {}
+        if filter_by: body['filter_by'] = request.filter_by.as_dict()
+        if include_metrics: body['include_metrics'] = request.include_metrics
+        if max_results: body['max_results'] = request.max_results
+        if page_token: body['page_token'] = request.page_token
 
+        # Only the first request can include the filter and params need to be in the body.
+        # Ref: https://github.com/databricks/databricks-sdk-py/issues/99
+        json = self._api.do('GET', '/api/2.0/sql/history/queries', body=body)
+        del body['filter_by']
         while True:
-            json = self._api.do('GET', '/api/2.0/sql/history/queries', query=query)
             if 'res' not in json or not json['res']:
                 return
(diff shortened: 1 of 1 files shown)


mgyucht:
Thanks for raising this PR @lukehsiao! Unfortunately we won't be able to accept this PR as this file is autogenerated and your change will be reverted on the next generation pass. We will need to change the program used to generate these clients to handle this case. You can see which files are auto-generated in our SDK by looking at the first line of the file. If you want to address this change, the right place to do it is actually in the ApiClient implementation itself. You can inspect the request endpoint and modify it as needed there. Feel free to give that a shot!

FYI: internally, we're working on fixing the handling of parameters for GET and DELETE APIs across the entire API surface of Databricks. Most endpoints do handle parameters in a compliant manner, but this endpoint doesn't yet. I'll double check on the progress of this. However, the issue with filter_by and next_page_token will still need separate handling.

lukehsiao:
> If you want to address this change, the right place to do it is actually in the ApiClient implementation itself. You can inspect the request endpoint and modify it as needed there. Feel free to give that a shot!

Apologies, can you say more here? I see the `ApiClient` class in `core.py`, but...it's not immediately obvious to me how that affects this file.

> internally, we're working on fixing the handling of parameters for GET and DELETE APIs across the entire API surface of Databricks. Most endpoints do handle parameters in a compliant manner, but this endpoint doesn't yet. I'll double check on the progress of this. However, the issue with filter_by and next_page_token will still need separate handling.

An update here would be great! We would like to rely on Query History via this SDK in our application and having this work in the SDK would be very helpful.

PR #248, opened 2023-07-21
Call .value when serializing lists of enum values

### LINKED ISSUE #214: IAM: WorkspacePermission Enum value
Hi,

There is an issue in the following line (UpdateWorkspaceAssignments, line 787): https://github.com/databricks/databricks-sdk-py/blob/befbb42329a42727c530b6ac8101e0ff20be5013/databricks/sdk/service/iam.py#L787C9-L787C81

WorkspacePermission should called with the Enum value:
```
if self.permissions: body['permissions'] = [v.value for v in self.permissions]
```

mgyucht:
## Changes
Currently, lists of enum values in request objects are incorrectly serialized in the `as_dict()` method. Instead of putting the enum directly into the resulting dictionary, the enum's value needs to be added to the dictionary.

Closes #214.

## Tests
Updated test_cluster_events to add a filter on EventType, which is a list of enum values.

### DIFF
--- a/.codegen/service.py.tmpl
+++ b/.codegen/service.py.tmpl
@@ -55,7 +55,7 @@ class {{.PascalName}}{{if eq "List" .PascalName}}Request{{end}}:{{if .Descriptio
 {{- end -}}
 {{- define "as_request_type" -}}
 	{{- if not .Entity }}None # ERROR: No Type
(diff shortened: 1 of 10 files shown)


tanmay-db (approved):
Was this not detected earlier on because there weren't any tests covering this?

mgyucht:
I don't think there is any test covering this now. The test I added failed when run on main but passes on my branch.

PR #262, opened 2023-08-03
Removed `**kwargs` from methods

nfx:
This PR removes `XxxRequest` classes that are referenced only as requests, effectively removing the undocumented `request=(XxxRequest(foo=1))` API. This will enforce validation on the field names and make error messages more clear.


mgyucht:
Can we separate these two changes into two separate PRs?

nfx:
integration tests pass

image

mgyucht (approved):
LGTM!

PR #276, opened 2023-08-08
Added `connection_pool_size` configuration property (preview)

nfx:
We have three things that are configurable from `requests` connection pooling perspective:

* `pool_connections` - Number of urllib3 connection pools to cache before discarding the least recently used pool. Python requests default value is 10. This PR increases it to 20.
* `pool_maxsize` - The maximum number of connections to save in the pool. Improves performance in multithreaded situations. For now, we're setting it to the same value as connection_pool_size.
* `pool_block` - If pool_block is False, then more connections will are created, but not saved after the first use. Block when no free connections are available. urllib3 ensures that no more than pool_maxsize connections are used at a time. Prevents platform from flooding. By default, requests library doesn't block. We start with blocking.

This PR introduces two new configuration properties:
- `connection_pool_size` configuration property, which sets `pool_connections`. Defaults to 20.
- `connection_pool_max_size`, which sets `pool_maxsize`. Defaults to the same value as `connection_pool_size`.

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -491,8 +491,8 @@ class Config:
     metadata_service_url = ConfigAttribute(env='DATABRICKS_METADATA_SERVICE_URL',
                                            auth='metadata-service',
                                            sensitive=True)
-    connection_pool_size: int = ConfigAttribute()
-    connection_pool_max_size: int = ConfigAttribute()
+    max_connection_pools: int = ConfigAttribute()
+    max_connections_per_pool: int = ConfigAttribute()
 
     def __init__(self,
                  *,
@@ -898,15 +898,15 @@ class ApiClient:
 
         # Number of urllib3 connection pools to cache before discarding the least
         # recently used pool. Python requests default value is 10.
-        pool_connections = cfg.connection_pool_size
+        pool_connections = cfg.max_connection_pools
         if pool_connections is None:
             pool_connections = 20
 
         # The maximum number of connections to save in the pool. Improves performance
         # in multithreaded situations. For now, we're setting it to the same value
         # as connection_pool_size.
-        pool_maxsize = cfg.connection_pool_max_size
-        if cfg.connection_pool_max_size is None:
+        pool_maxsize = cfg.max_connections_per_pool
+        if cfg.max_connections_per_pool is None:
             pool_maxsize = pool_connections
 
         # If pool_block is False, then more connections will are created,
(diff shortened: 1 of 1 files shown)


databricks/sdk/core.py
```diff
@@ -893,7 +894,29 @@ def __init__(self, cfg: Config = None):
 
         self._session = requests.Session()
         self._session.auth = self._authenticate
-        self._session.mount("https://", HTTPAdapter(max_retries=retry_strategy))
+
+        # Number of urllib3 connection pools to cache before discarding the least
+        # recently used pool. Python requests default value is 10.
+        pool_connections = cfg.connection_pool_size
+        if pool_connections is None:
+            pool_connections = 20
```
mgyucht, line 902: I understand the need to increase the connection pool size to support concurrent requests. However, why do we need multiple connection pools? Looking at `requests`, HttpAdapter's underlying PoolManager uses the same pool based on the <host,port,scheme> tuple, which for a single client will only be one endpoint.

mgyucht (commented):
+1 on increasing connection pool size & blocking. However, I don't know about increasing # of connection pools.

databricks/sdk/core.py
nfx: I also do think that we need just one pool, but somehow the default client was using the same value for both.
how about exposing two configuration properties, marking them experimental, and figure out the best defaults later? single SDK client will query only one host.

how should we name `pool_maxsize`? `connection_pool_max`?

mgyucht: Default uses the same value for both because it uses the same constant for each parameter as default values in its constructor:
```
    def __init__(
        self,
        pool_connections=DEFAULT_POOLSIZE,
        pool_maxsize=DEFAULT_POOLSIZE,
        max_retries=DEFAULT_RETRIES,
        pool_block=DEFAULT_POOLBLOCK,
    ):
```
`DEFAULT_POOLSIZE` is set to 10.

databricks/sdk/core.py
```diff
@@ -491,6 +491,8 @@ class Config:
     metadata_service_url = ConfigAttribute(env='DATABRICKS_METADATA_SERVICE_URL',
                                            auth='metadata-service',
                                            sensitive=True)
+    connection_pool_size: int = ConfigAttribute()
+    connection_pool_max_size: int = ConfigAttribute()
```
mgyucht, line 495: Here's my suggestion for the naming:
* `max_connection_pools`: maximum number of connection pools used by the HttpAdapter.
* `max_connections_per_pool`: maximum number of connections per pool used by the HttpAdapter.

mgyucht (approved):
Small nit on naming, but otherwise the change seems fine.

databricks/sdk/core.py
nfx: sure. let's get it shipped

PR #283, opened 2023-08-11
Retry in `w.clusters.ensure_cluster_is_running(id)` when cluster is simultaneously started by two different processes.

nfx:
This PR adds a retry for timing edge cases like `INVALID_STATE: Cluster XXX is in unexpected state Pending.`

Other PRs:
- https://github.com/databricks/databricks-sdk-go/pull/580
- https://github.com/databricks/databricks-sdk-java/pull/134

### DIFF
--- a/databricks/sdk/mixins/compute.py
+++ b/databricks/sdk/mixins/compute.py
@@ -5,6 +5,7 @@ import time
 from dataclasses import dataclass
 from typing import Optional
 
+from databricks.sdk.core import DatabricksError
 from databricks.sdk.errors import OperationFailed
 from databricks.sdk.service import compute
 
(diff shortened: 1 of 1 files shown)


databricks/sdk/mixins/compute.py
```diff
@@ -231,6 +232,11 @@ def ensure_cluster_is_running(self, cluster_id: str) -> None:
                     return
                 elif info.state in (state.ERROR, state.UNKNOWN):
                     raise RuntimeError(f'Cluster {info.cluster_name} is {info.state}: {info.state_message}')
+            except DatabricksError as e:
+                if e.error_code == 'INVALID_STATE':
+                    _LOG.debug(f'Cluster was started by other process: {e} Retrying.')
+                    continue
```
igorsmi-db, line 238: It might make sense to retry only once/limited number of times. Because who knows what are the other errors can go from cluster-manager with this error code. And if there is something else, it will be an infinite loop

nfx: we retry this loop with a timeout, look at the beginning of the method

mgyucht (approved):
LGTM!

PR #284, opened 2023-08-11
(PoC) Parallel recursive listing for `w.workspace.list()`

nfx:


databricks/sdk/mixins/workspace.py
```diff
@@ -8,6 +10,54 @@
...
+class _ParallelRecursiveListing:
+    def __init__(self, path, listing, threads, notebooks_modified_after):
+        self.path = path
+        self.listing = listing
+        self.threads = threads
+        self.notebooks_modified_after = notebooks_modified_after
+        self.directories = Queue()
+        self.results = Queue()
+        self.directories.put_nowait(path)
+        self._start()
```
pietern, line 24: Recommend against starting from the constructor, can start from `__iter__`.

databricks/sdk/mixins/workspace.py
```diff
@@ -8,6 +10,54 @@
...
+        while True:
+            path = self.directories.get()
+            if path is None:
+                _LOG.debug('stopping thread')
+                break # poison pill
+            for object_info in self.listing(
+                    path, notebooks_modified_after=self.notebooks_modified_after):
+                if object_info.object_type == ObjectType.DIRECTORY:
+                    self.directories.put(object_info.path)
+                    continue
```
pietern, line 36: `put` vs `put_nowait` doesn't make a difference for an unbounded queue.

databricks/sdk/mixins/workspace.py
```diff
@@ -8,6 +10,54 @@
...
+                _LOG.debug(f'found: {object_info.path}')
+                self.results.put_nowait(object_info)
+            self.directories.task_done()
```
pietern, line 39: This is not used; you're not using `Queue.join` to wait for completion.

databricks/sdk/mixins/workspace.py
```diff
@@ -8,6 +10,54 @@
...
+        import concurrent.futures
+        with concurrent.futures.ThreadPoolExecutor(max_workers=self.threads) as pool:
+            tasks = []
+            for _ in range(self.threads):
+                tasks.append(pool.submit(self._worker))
+            concurrent.futures.wait(tasks)
+
+    def __iter__(self) -> Iterator[ObjectInfo]:
+
+        while self.results.not_empty:
```
pietern, line 55: This can race; you need to also check if the queue is empty, and do so atomically.

pietern (commented):
Couple notes, I think this probably hangs as-is.

databricks/sdk/mixins/workspace.py
nfx: yep, it's a PoC still

databricks/sdk/mixins/workspace.py
nfx: results queue can be empty for short periods of time and then full again. will add "is_running" status, that it flipped to false before sending poison pills to workers

databricks/sdk/mixins/workspace.py
```diff
@@ -1,14 +1,128 @@
...
+                    raise err
+                _LOG.warning(f"{path} is not listable. Ignoring")
+            finally:
+                self._leave_folder(path)
+        _LOG.debug(f"Finished workspace listing worker {num}")
+
+    def _list(self, path):
+        listing = self._listing(path, notebooks_modified_after=self._notebooks_modified_after)
+        for object_info in sorted(listing, key=lambda _: _.path):
+            if object_info.object_type != ObjectType.DIRECTORY:
```
mgyucht, line 100: I think this should be `==`? The code in this block refers to what happens in a directory.

databricks/sdk/mixins/workspace.py
```diff
@@ -1,14 +1,128 @@
...
+            for _ in range(0, self._threads - 1):
+                self._work.put(None)
+            with self._reporter_cond:
+                self._reporter_cond.notify()
+
+    def _is_running(self):
+        with self._cond:
+            return self._in_progress > 0
+
+    def _reporter(self):
```
mgyucht, line 68: The reporter is a bit weird. Given that this is a generator and users may not list all of the items "right away", this would just continue to print logs until the generator is consumed. Alternatively, this could be provided as a parameter to the parallel list operation, allowing users to provide a callback that is periodically called with statistics like this if they desire, then they can control when/whether they want to log or do something else.

mgyucht (commented):
I think this is a good addition to the SDK. Let's drop the reporter for now, or if you want, add it as a parameter to this API. Aside from this, I think there is one bug in the `_list` method.

Can we add an integration test where we actually create some recursive directory structure, try to list everything, and check that it worked correctly?

renaudhartert-db:

PR #287, opened 2023-08-15
Added collection of Databricks Runtime versions used together with Python SDK

nfx:
This PR adds collection of DBR version if used from within Databricks Cluster. This will allow to improve on notebook-native auth.

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -614,6 +614,10 @@ class Config:
             ua.append(' '.join(self._user_agent_other_info))
         if len(self._upstream_user_agent) > 0:
             ua.append(self._upstream_user_agent)
+        if 'DATABRICKS_RUNTIME_VERSION' in os.environ:
+            runtime_version = os.environ['DATABRICKS_RUNTIME_VERSION']
+            if runtime_version != '':
+                ua.append(f'runtime/{runtime_version}')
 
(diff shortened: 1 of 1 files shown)


databricks/sdk/core.py
```diff
@@ -614,6 +614,10 @@ def user_agent(self):
             ua.append(' '.join(self._user_agent_other_info))
         if len(self._upstream_user_agent) > 0:
             ua.append(self._upstream_user_agent)
+        if 'DATABRICKS_RUNTIME_VERSION' in os.environ:
+            runtime_version = os.environ['DATABRICKS_RUNTIME_VERSION']
+            if runtime_version != '':
+                ua.append(f'runtime/{runtime_version}')
```
pietern, line 620: Please add a check that this value conforms to an expected pattern, e.g. no spaces/slashes.

Can be a simple regex match. 

I suspect you expect to see `13.3.x-scala2.12`?

nfx: @pietern it's `13.3`

mgyucht (approved):
Nice, thanks!

mgyucht:
As no good deed goes unpunished, can you make this change in Go/Java as well?

mgyucht:
Actually, I guess it is not needed in Go, but Java would.

PR #302, opened 2023-08-19
Propagate request headers to ApiClient

mgyucht:
## Changes
Based on changes in https://github.com/databricks/databricks-sdk-go/pull/572.

This PR adds initial support for headers in the ApiClient class. When making a request, the SDK will construct a map of headers to be included in the request. These are passed through a new parameter in the `do` method.

In the future, to support dynamic headers, we can modify our codegen template to add those fields to the headers map before calling the ApiClient.

Alternatives considered:

* **Adding headers in request objects directly**: this would not require any change in the ApiClient interface. Instead, we would introduce a new annotation for headers, much like what we do for query parameters, and reflectively scan the fields of each request object. We could do this, but for headers that have a fixed value (like Content-Type and Accept), these fields would essentially be private final fields that we could iterate through. Users can never interact with these fields, so it seems a bit unusual to have them in the request structure, a public interface.
## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


.codegen/service.py.tmpl
```diff
@@ -202,6 +203,12 @@ class {{.Name}}API:{{if .Description}}
         {{- end}}
 {{- end}}
 
+{{ define "method-headers" -}}
```
nfx, line 206: please surround it with `if`

.codegen/service.py.tmpl
```diff
@@ -278,9 +285,15 @@ class {{.Name}}API:{{if .Description}}
...
-{{- else}}'{{.Path}}'{{end}}{{if .Request}}{{if .Request.HasQueryField}}, query=query{{end}}{{if .Request.HasJsonField}}, body=body{{end}}{{end}})
+self._api.do('{{.Verb}}',
+  {{if .PathParts -}}
+      f'{{range  .PathParts}}{{.Prefix}}{{if .Field}}{{"{"}}{{.Field.SnakeName}}{{with .Field.Entity.Enum}}.value{{end}}{{"}"}}{{else if .IsAccountId}}{{"{self._api.account_id}"}}{{end}}{{ end }}'
+  {{- else}}'{{.Path}}'{{end}}
+  {{if .Request}}
+      {{- if .Request.HasQueryField}}, query=query{{end}}
+      {{- if .Request.HasJsonField}}, body=body{{end}}
+  {{end}}
+  , headers=headers)
```
nfx, line 296:
```suggestion
  {{if .HasHeaders}}, headers=headers{{end}})
```

databricks/sdk/core.py
```diff
@@ -981,11 +981,12 @@ def do(self,
            method: str,
            path: str,
            query: dict = None,
+           headers: dict = None,
            body: dict = None,
            raw: bool = False,
            files=None,
            data=None) -> dict:
-        headers = {'Accept': 'application/json', 'User-Agent': self._user_agent_base}
+        headers |= {'User-Agent': self._user_agent_base}
```
nfx, line 989: a) may not work in python 3.7
b) this will crash, as headers default value is empty

nfx (dismissed):
the diff seems to be larger than required. e.g. in 500 methods, the `'Accept': 'application/json'` and `'Content-Type': 'application/json'` can be defaults and hidden within `ApiClient#do`.

tanmay-db (approved):
LGTM

PR #303, opened 2023-08-19
Support BinaryIO for streaming request and response bodies

mgyucht:
## Changes
This PR adds support for non-application/json requests and responses. This PR is divided into two parts:

**Support setting headers per request**. The `do` method of the ApiClient class is expanded to accept one new parameter, `headers: dict`, and all callers are expected to pass a map of headers to be included in the request. As the allowed content types for requests and responses are known from the OpenAPI specification, callers must construct the request header map in code generation and pass it to the ApiClient. The default "Content-Type: application/json" header is removed, as each request should specify its own Content-Type and Accept headers. This is done in https://github.com/databricks/databricks-sdk-py/pull/302.

**Add support for streaming request and response bodies to support non-application/json requests/responses**. Today, serialization of requests is done in the ApiClient. This implies that ApiClient needs to be able to serialize a request purely based on the parameters provided to it via headers, the request body, etc. In this proposal, the serialization is specified by the caller depending on whether the `body` or `data` field of `do()` is used: `data` should be used for already-serialized data, such as BinaryIO or file-like objects, and `body` should be used for JSON objects.

One important caveat about streaming responses is that callers are required to close the streams when they are done using them. To do this, we may need to expose the raw Response object so that users can use `with w.service.method() as response:` to automatically close the response when they are done reading.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


.codegen/service.py.tmpl
```diff
@@ -3,8 +3,9 @@
 from dataclasses import dataclass
 from datetime import timedelta
 from enum import Enum
-from typing import Dict, List, Any, Iterator, Type, Callable, Optional
+from typing import Dict, List, Any, Iterator, Type, Callable, Optional, BinaryIO
 import time
+import io
```
mgyucht, line 8: Remove this?

.codegen/service.py.tmpl
```diff
@@ -195,11 +198,13 @@ class {{.Name}}API:{{if .Description}}
...
-        query = {}{{end}}
-        {{- range .Request.Fields}}{{if not .IsPath}}
-        if {{.SnakeName}} is not None: {{if .IsQuery}}query{{else}}body{{end}}['{{.Name}}'] = {{template "method-param-bind" .}}{{end}}{{end}}
+        {{if .Request.HasJsonField -}}
+        body = {}{{ end }}{{if .Request.HasQueryField}}
+        query = {}{{ end }}{{ range .Request.Fields }}{{ if not .IsPath -}}
+        {{ if .IsJson }}
+        if {{.SnakeName}} is not None: body['{{.Name}}'] = {{template "method-param-bind" .}}{{end}}{{ if .IsQuery }}
+        if {{.SnakeName}} is not None: query['{{.Name}}'] = {{template "method-param-bind" .}}{{end}}
+        {{- end}}
```
mgyucht, line 207: Can I safely revert this?

tanmay-db (approved):
Went over the PR with Miles, LGTM

PR #306, opened 2023-08-21
[DECO-2229] Handle GZIP'ed streaming responses

hectorcast-db:
## Changes
Handle GZIP'ed streaming responses by using iter_content method instead of the raw response.

Relevant documentation:
[https://requests.readthedocs.io/en/latest/user/quickstart](https://requests.readthedocs.io/en/latest/user/quickstart/#:~:text=iter_content%20will%20automatically%20decode%20the,they%20were%20returned%2C%20use%20Response)

## Tests

- [X] `make test` run locally
- [X] `make fmt` applied
- [X] relevant integration tests applied: `test_files.py` and `test_workspace.py`

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -15,8 +15,8 @@ import urllib.parse
 from datetime import datetime
 from json import JSONDecodeError
 from types import TracebackType
-from typing import (Any, BinaryIO, Callable, Dict, Iterable, Iterator,
-                    List, Optional, Type, Union)
+from typing import (Any, BinaryIO, Callable, Dict, Iterable, Iterator, List,
+                    Optional, Type, Union)
 
 import requests
 import requests.auth
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -14,7 +14,9 @@ import sys
 import urllib.parse
 from datetime import datetime
 from json import JSONDecodeError
-from typing import Any, Callable, Dict, Iterable, List, Optional, Union
+from types import TracebackType
+from typing import (Any, AnyStr, BinaryIO, Callable, Dict, Iterable, Iterator,
+                    List, Optional, Type, Union)
 
 import requests
 import requests.auth
@@ -984,7 +986,7 @@ class ApiClient:
            body: dict = None,
            raw: bool = False,
            files=None,
-           data=None) -> dict:
+           data=None) -> Union[dict, BinaryIO]:
         headers = {'Accept': 'application/json', 'User-Agent': self._user_agent_base}
         response = self._session.request(method,
                                          f"{self._cfg.host}{path}",
@@ -1002,7 +1004,7 @@ class ApiClient:
                 payload = response.json()
                 raise self._make_nicer_error(status_code=response.status_code, **payload) from None
             if raw:
-                return response.raw
+                return StreamingResponse(response)
             if not len(response.content):
                 return {}
             return response.json()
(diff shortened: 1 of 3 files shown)


tests/integration/test_workspace.py
```diff
@@ -21,6 +21,22 @@ def test_workspace_upload_download_notebooks(w, random):
     w.workspace.delete(notebook)
 
 
+def test_workspace_unzip_notebooks(w, random):
```
mgyucht, line 24: Can we add a test to download the notebook 100 times and make sure that we don't exhaust our connection pool? We default to 20, so let's do 30 or so to be safe.

tests/integration/test_workspace.py
```diff
@@ -21,6 +21,22 @@ def test_workspace_upload_download_notebooks(w, random):
...
+    notebook = f'/Users/{w.current_user.me().user_name}/notebook-{random(12)}.py'
+
+    # Big notebooks can be gzipped during transfer by the API (out of our control)
+    # Creating some large content to trigger this behaviour
+    notebook_content = ('print(1)\n' * 1000).strip('\n')
```
mgyucht, line 29: Nice notebook :)

databricks/sdk/core.py
```diff
@@ -1118,3 +1120,82 @@ def _redacted_dump(self, prefix: str, body: str) -> str:
...
+    def writable(self) -> bool:
+        return False
+
+    def write(self, s: AnyStr) -> int:
+        raise NotImplementedError()
+
+    def writelines(self, lines: Iterable[AnyStr]) -> None:
+        raise NotImplementedError()
+
+    def __next__(self) -> AnyStr:
```
mgyucht, line 1191: Just looking at this now: this should probably return `self.read(1)`?

tests/integration/test_workspace.py
mgyucht, line 24: Can we add a test that calls `read()` with a parameter to test reading `N` bytes?

databricks/sdk/core.py
```diff
@@ -1118,3 +1120,82 @@ def _redacted_dump(self, prefix: str, body: str) -> str:
             return "\n".join([f'{prefix}{line}' for line in json.dumps(raw, indent=2).split("\n")])
         except JSONDecodeError:
             return f'{prefix}[non-JSON document of {len(body)} bytes]'
+
+
+class StreamingResponse(BinaryIO):
+    _response: requests.Response
+    _buffer: bytes
+    _content: Iterator[any]
```
mgyucht, line 1128: I think this should be `Iterator[bytes]`: https://github.com/urllib3/urllib3/blob/1adddfd93ba0f880d74652ca7fe4ca69160ff2d5/src/urllib3/response.py#L381

databricks/sdk/core.py
```diff
@@ -1118,3 +1120,82 @@ def _redacted_dump(self, prefix: str, body: str) -> str:
...
+        self._content = self._response.iter_content()
+        return self
+
+    def close(self) -> None:
+        self._response.close()
+
+    def isatty(self) -> bool:
+        return False
+
+    def read(self, n: int = -1) -> AnyStr:
```
mgyucht, line 1144: Based on my reading of `AnyStr`, I think we should be a bit more narrow in terms of how we annotate this type. IMO `bytes` is more accurate. Looking at BinaryIO in typing.py:
```
class BinaryIO(IO[bytes]):
    ...
```
I think we can replace all instances of `AnyStr` with `bytes` in this class.

mgyucht (commented):
Changes mostly look good to me. Couple questions/notes, let's do one more pass and we can merge this tomorrow.

tests/integration/test_workspace.py
hectorcast-db: I will add that one.

hectorcast-db: Added as Unit Test

databricks/sdk/core.py
```diff
@@ -1118,3 +1120,85 @@ def _redacted_dump(self, prefix: str, body: str) -> str:
...
+    def write(self, s: AnyStr) -> int:
+        raise NotImplementedError()
+
+    def writelines(self, lines: Iterable[AnyStr]) -> None:
+        raise NotImplementedError()
+
+    def __next__(self) -> AnyStr:
+        return self.read(1)
+
+    def __iter__(self) -> Iterator[AnyStr]:
```
mgyucht, line 1197:
```suggestion
    def __iter__(self) -> Iterator[bytes]:
```

databricks/sdk/core.py
```diff
@@ -1118,3 +1120,85 @@ def _redacted_dump(self, prefix: str, body: str) -> str:
...
+    def truncate(self, __size: Union[int, None] = ...) -> int:
+        raise NotImplementedError()
+
+    def writable(self) -> bool:
+        return False
+
+    def write(self, s: AnyStr) -> int:
+        raise NotImplementedError()
+
+    def writelines(self, lines: Iterable[AnyStr]) -> None:
```
mgyucht, line 1191:
```suggestion
    def writelines(self, lines: Iterable[bytes]) -> None:
```

databricks/sdk/core.py
```diff
@@ -1118,3 +1120,85 @@ def _redacted_dump(self, prefix: str, body: str) -> str:
...
+            remaining_bytes -= to_read
+        return res
+
+    def readable(self) -> bool:
+        return self._content is not None
+
+    def readline(self, __limit: int = ...) -> AnyStr:
+        raise NotImplementedError()
+
+    def readlines(self, __hint: int = ...) -> List[AnyStr]:
```
mgyucht, line 1170:
```suggestion
    def readlines(self, __hint: int = ...) -> List[bytes]:
```

databricks/sdk/core.py
```diff
@@ -1118,3 +1120,85 @@ def _redacted_dump(self, prefix: str, body: str) -> str:
...
+            to_read = bytes_available if read_everything else min(remaining_bytes, bytes_available)
+            res += self._buffer[:to_read]
+            self._buffer = self._buffer[to_read:]
+            remaining_bytes -= to_read
+        return res
+
+    def readable(self) -> bool:
+        return self._content is not None
+
+    def readline(self, __limit: int = ...) -> AnyStr:
```
mgyucht, line 1167:
```suggestion
    def readline(self, __limit: int = ...) -> bytes:
```

databricks/sdk/core.py
```diff
@@ -1118,3 +1120,85 @@ def _redacted_dump(self, prefix: str, body: str) -> str:
...
+    def tell(self) -> int:
+        raise NotImplementedError()
+
+    def truncate(self, __size: Union[int, None] = ...) -> int:
+        raise NotImplementedError()
+
+    def writable(self) -> bool:
+        return False
+
+    def write(self, s: AnyStr) -> int:
```
mgyucht, line 1188:
```suggestion
    def write(self, s: bytes) -> int:
```

databricks/sdk/core.py
```diff
@@ -1118,3 +1120,85 @@ def _redacted_dump(self, prefix: str, body: str) -> str:
...
+        raise NotImplementedError()
+
+    def writelines(self, lines: Iterable[AnyStr]) -> None:
+        raise NotImplementedError()
+
+    def __next__(self) -> AnyStr:
```
mgyucht, line 1194:
```suggestion
    def __next__(self) -> bytes:
```

mgyucht (approved):
LGTM! Great tests. Let's replace AnyStr with bytes everywhere as per my suggestions.

PR #312, opened 2023-08-28
Check workspaceUrl explicitly in runtime repl auth

mgyucht:
## Changes
In runtime repl auth, it can be the case that the returned context object is not None, but the workspaceUrl is None. This results in the SDK being configured to use a workspace with URL `https://None`. This check allows runtime native auth to fall back to "legacy" auth.

## Tests
- [x] Built the library, uploaded to a 12.2 LTS cluster in GCP, listed clusters.

- [ ] `make test` run locally
- [ ] `make fmt` applied


hectorcast-db (approved):

nfx:
What data security mode does it happen on? Is it interactive or job?

mgyucht:
I tested this on an interactive single-user cluster.

PR #319, opened 2023-08-30
Improve file download performance

mgyucht:
## Changes
The change to use a chunked iterator for streaming responses caused a performance regression when streaming raw responses, as the default chunk size from the requests library is 1 byte. To identify an appropriate chunk size to use, I added a benchmark integration test that uploads a 50mb file and attempts to download it with a number of various chunk sizes. Based on this benchmarking, it seems like we don't have any real speed-up after around 100KiB chunks.

Additionally, I added a check here to ensure that `read()` raises a ValueError if the response was ever closed, as it is not allowed to read multiple times from a streaming response.

## Tests
- [x] Added an integration test to verify `read()` raises ValueError on second attempt to read.
- [x] Added a benchmark to stress-test file download performance.

```
[chunk size 1kb] Average time to download: 129.83376684188843
[chunk size 2kb] Average time to download: 69.17960963249206
[chunk size 5kb] Average time to download: 34.08896443843842
[chunk size 10kb] Average time to download: 19.392758226394655
[chunk size 20kb] Average time to download: 10.74389090538025
[chunk size 50kb] Average time to download: 5.658655261993408
[chunk size 100kb] Average time to download: 3.982270860671997
[chunk size 200kb] Average time to download: 4.485624170303344
[chunk size 500kb] Average time to download: 4.236340761184692
[chunk size 1000kb] Average time to download: 4.695496129989624
[chunk size 2000kb] Average time to download: 4.6709349155426025
[chunk size 5000kb] Average time to download: 4.5816244840621945
[chunk size 10000kb] Average time to download: 4.32698233127594
[chunk size 20000kb] Average time to download: 4.625458240509033
[chunk size 50000kb] Average time to download: 4.405146503448487
{1: 129.83376684188843, 2: 69.17960963249206, 5: 34.08896443843842, 10: 19.392758226394655, 20: 10.74389090538025, 50: 5.658655261993408, 100: 3.982270860671997, 200: 4.485624170303344, 500: 4.236340761184692, 1000: 4.695496129989624, 2000: 4.6709349155426025, 5000: 4.5816244840621945, 10000: 4.32698233127594, 20000: 4.625458240509033, 50000: 4.405146503448487}
Fastest chunk size:  100 kb,  3.982270860671997 seconds
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ integration:
 	pytest -n auto -m 'integration and not benchmark' --cov=databricks --cov-report html tests
 
 benchmark:
-	pytest -m 'benchmark' --cov=databricks --cov-report html tests
+	pytest -m 'benchmark' tests
 
 coverage: test
 	open htmlcov/index.html


pietern (approved):
It depends on the bandwidth/latency of the link as well. Ultimately the socket buffer on the read side needs to be drained fast enough for the TCP buffers to keep filling.

From the requests docs:
> chunk_size must be of type int or None. A value of None will function differently depending on the value of stream. stream=True will read data as it arrives in whatever size the chunks are received. If stream=False, data is returned as a single chunk.

Did you try passing chunk size equal to `None` as well?

mgyucht:
I did not try that out. Let me add that to the benchmark. I think `None` makes sense based on my reading: as soon as bytes are available to be read, we can make them available to users. I'll add it to my benchmark and repeat.

mgyucht:
New results are:
```
{20: 13.47412874698639, 50: 8.377137064933777, 100: 7.477979683876038, 200: 6.942076921463013, 500: 8.758025193214417, 1000: 6.48618483543396, 2000: 6.075670170783996, 5000: 6.664759802818298, 10000: 6.4036040782928465, 20000: 6.440246605873108, 50000: 5.960023260116577, None: 5.902777791023254}
```
It seems like it levels off around 200 KiB this time, but `None` has the edge over everything, so I'll go with that.

PR #324, opened 2023-08-30
Fixed repos.list() example in docs.

camlee:
## Changes
Simple docs change to fix an example that doesn't work: the `repos.list()` function doesn't take any positional arguments.

## Tests
No testing done as this is a docs change.

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/docs/workspace/repos.rst
+++ b/docs/workspace/repos.rst
@@ -129,7 +129,7 @@ Repos
             
             w = WorkspaceClient()
             
-            all = w.repos.list(workspace.ListReposRequest())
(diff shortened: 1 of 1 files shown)


camlee:
Ahhh. Is that generated from the examples folder here: https://github.com/databricks/databricks-sdk-py/blob/3bfe1d08b628a1a60a38f2bedbad92af9817dca7/examples/repos/list_repos.py#

If yes, would a proper fix bo to change it there? And presumably re-generating the docs would then sync the two places up.

PR #326, opened 2023-08-31
Fix test_files_api_read_twice_from_one_download

mgyucht:
## Changes
This integration test is failing (and likely wasn't passing when I merged the original PR). This PR fixes it by unifying the codepaths for `__enter__` and `read` to call `_open`, which consistently will check if the response is closed and throw an exception if so.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -1147,8 +1147,14 @@ class StreamingResponse(BinaryIO):
         self._content = None
         self._chunk_size = chunk_size
 
+    def _open(self) -> None:
+        if self._closed:
+            raise ValueError("I/O operation on closed file")
+        if not self._content:
+            self._content = self._response.iter_content(chunk_size=self._chunk_size)
+
     def __enter__(self) -> BinaryIO:
-        self._content = self._response.iter_content(chunk_size=self._chunk_size)
+        self._open()
         return self
 
     def set_chunk_size(self, chunk_size: Union[int, None]) -> None:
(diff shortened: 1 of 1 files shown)


databricks/sdk/core.py
```diff
@@ -1147,8 +1147,14 @@ def __init__(self, response: requests.Response, chunk_size: Union[int, None] = N
         self._content = None
         self._chunk_size = chunk_size
 
+    def _open(self) -> None:
+        if self._closed:
+            raise ValueError("I/O operation on closed file")
+        if not self._content:
+            self._content = self._response.iter_content(chunk_size=self._chunk_size)
```
pietern, line 1154: If the chunk size is passed in the constructor, this can also be done from the constructor and the need for a separate function (both `set_chunk_size` and `_open`) is gone.

pietern (approved):

databricks/sdk/core.py
mgyucht: Today the chunk size is not passed to the constructor in the `do()` method of `Client` because this would imply that there is a way to set it at construction time. This would require either a client-wide chunk size default or exposing a parameter in the `do()` method to set the chunk size. I felt this way was easier where we can have the default/preferred behavior but can override it as needed without having to feed parameters for the constructor through the client (and generated code).

pietern: Makes sense.

Remaining question; why _can_ it be set in the constructor if it is not used?

mgyucht: Good point. It's not strictly necessary. I kind of feel like it's a good practice to be able to initialize your object the way you want to purely from the constructor if needed rather than having to do
```
x = StreamingResponse(raw)
x.set_chunk_size(123)
return x
```

PR #327, opened 2023-08-31
Don't try to import runtime_auth when not in runtime

kartikgupta-db:
## Changes
Runtime auth is attempted when we do not have any config variables set (either directly in the sdk config, in .databrickcfg or as env vars). This further attempts importing runtime_auth functions from the dbruntime package. These are not available outside dbr. 

This PR moves "is in dbr runtime" check before the import, so that the import is not attempted outside dbr and users don't get cryptic errors. 
## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


pietern (approved):
Thanks!

databricks/sdk/core.py
```diff
@@ -94,11 +94,16 @@ def inner() -> Dict[str, str]:
 
 @credentials_provider('runtime', [])
 def runtime_native_auth(cfg: 'Config') -> Optional[HeaderFactory]:
+    if 'DATABRICKS_RUNTIME_VERSION' not in os.environ:
+        logging.debug("Skipping runtime authentication, because DATABRICKS_RUNTIME_VERSION is not set")
```
nfx, line 98:
```suggestion
```

This logline would appear in 75% of deployments, raising suspicions. let's remove it.

nfx (changes requested):

databricks/sdk/core.py
kartikgupta-db: @mgyucht removing this line to get this PR merged. We can discuss adding this back in when you are back.

nfx (approved):

PR #328, opened 2023-08-31
[DECO-2485] Handle Azure authentication when WorkspaceResourceID is provided

hectorcast-db:
## Changes
Handle Azure authentication when WorkspaceResourceID is provided

Get token for the correct subscription 

## Tests
* Created Unit tests
* Manually listed workspace cluster in the following scenarios:
  * User with wrong default tenant. No WorkspaceResourceID provided: Fail (expected). WARN log emitted.
  * User with wrong default tenant. WorkspaceResourceID provided: Succeed
  * User with no subscription. No WorkspaceResourceID provided: Succeed. WARN log  emitted.
  * User with no subscription. WorkspaceResourceID provided: Succeed (fallback mode, expected).

- [X] `make test` passing
- [X] `make fmt` applied
https://github.com/databricks/eng-dev-ecosystem/actions/runs/6038981442

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -253,8 +253,11 @@ class CliTokenSource(Refreshable):
 class AzureCliTokenSource(CliTokenSource):
     """ Obtain the token granted by `az login` CLI command """
 
-    def __init__(self, resource: str):
+    def __init__(self, resource: str, subscription: str = ""):
         cmd = ["az", "account", "get-access-token", "--resource", resource, "--output", "json"]
+        if subscription != "":
+            cmd.append("--subscription")
+            cmd.append(subscription)
         super().__init__(cmd=cmd,
                          token_type_field='tokenType',
                          access_token_field='accessToken',
@@ -264,8 +267,8 @@ class AzureCliTokenSource(CliTokenSource):
 @credentials_provider('azure-cli', ['is_azure'])
 def azure_cli(cfg: 'Config') -> Optional[HeaderFactory]:
     """ Adds refreshed OAuth token granted by `az login` command to every request. """
-    token_source = AzureCliTokenSource(cfg.effective_azure_login_app_id)
-    mgmt_token_source = AzureCliTokenSource(cfg.arm_environment.service_management_endpoint)
+    token_source = get_token(cfg, cfg.effective_azure_login_app_id)
+    mgmt_token_source = get_token(cfg, cfg.arm_environment.service_management_endpoint)
     try:
         token_source.token()
     except FileNotFoundError:
@@ -278,7 +281,7 @@ def azure_cli(cfg: 'Config') -> Optional[HeaderFactory]:
         logger.debug(f'Not including service management token in headers', exc_info=e)
         mgmt_token_source = None
 
-    _ensure_host_present(cfg, lambda resource: AzureCliTokenSource(resource))
+    _ensure_host_present(cfg, lambda resource: get_token(cfg, resource))
     logger.info("Using Azure CLI authentication with AAD tokens")
 
     def inner() -> Dict[str, str]:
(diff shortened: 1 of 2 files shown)


databricks/sdk/core.py
```diff
@@ -292,6 +295,37 @@ def inner() -> Dict[str, str]:
     return inner
 
 
+def get_token(cfg: 'Config', resource: str) -> AzureCliTokenSource:
```
nfx, line 298: why can't this validation be moved to `AzureCliTokenSource` constructor?

databricks/sdk/core.py
```diff
@@ -292,6 +295,37 @@ def inner() -> Dict[str, str]:
...
+    subscription = get_subscription(cfg)
+    if subscription != "":
+        token = AzureCliTokenSource(resource, subscription)
+        try:
+            # This will fail if the user has access to the workspace, but not to the subscription
+            # itself.
+            # In such case, we fall back to not using the subscription.
+            token.token()
+            return token
+        except Exception:
```
nfx, line 308: don't catch broad exceptions, catch concrete ones.

databricks/sdk/core.py
```diff
@@ -292,6 +295,37 @@ def inner() -> Dict[str, str]:
...
+            logger.warning("Failed to get token for subscription. Using resource only token.")
+    else:
+        logger.warning(
+            "azure_workspace_resource_id field not provided. " +
+            "It is recommended to specify this field in the Databricks configuration to avoid authentication errors."
+        )
+    return AzureCliTokenSource(resource)
+
+
+def get_subscription(cfg: 'Config') -> str:
```
nfx, line 318: should move as a private method to `AzureCliTokenSource` as well

nfx (changes requested):

databricks/sdk/core.py
hectorcast-db: That would mean adding a try/except block around the "super().__init__" code. Not being that familiar with Python, and coming from languages like Java which don't allow you to do that, it seemed like a bad idea to me. But I can add it if it is considered acceptable practice in Python.

databricks/sdk/core.py
hectorcast-db: Changed

databricks/sdk/core.py
nfx: can you then try adding `validate` (or validate_subscription, or something like that)  method on `AzureCliTokenSource` and call it instead of the first call to `token`. 

```
    try:
        token_source.token()
    except FileNotFoundError:
```

nfx: btw, the `IncorrectClaimException` in the original ticket does refer to Azure **Tenant** (identities), not the Azure **Subscription** (resources). there are valid cases, when user has no access to subscription, but has access to a workspace. also, `azure_workspace_resource_id` is not required for Azure CLI auth type.

to solve `IncorrectClaimException` we have to supply `azure_login_app_id`. I'll reach to you on slack.

mgyucht, line 298:
```python
@staticmethod
def for_resource(cfg: 'Config', resource: str) -> 'AzureCliTokenSource':
    ...

AzureCliTokenSource.for_resource(...)
```

databricks/sdk/core.py
```diff
@@ -264,8 +267,8 @@ def __init__(self, resource: str):
 @credentials_provider('azure-cli', ['is_azure'])
 def azure_cli(cfg: 'Config') -> Optional[HeaderFactory]:
     """ Adds refreshed OAuth token granted by `az login` command to every request. """
-    token_source = AzureCliTokenSource(cfg.effective_azure_login_app_id)
-    mgmt_token_source = AzureCliTokenSource(cfg.arm_environment.service_management_endpoint)
+    token_source = get_token(cfg, cfg.effective_azure_login_app_id)
```
mgyucht, line 270: Move error-handling logic to `for_resource`

mgyucht (commented):
As discussed, need a couple changes to error handling, and the new functions should be internal to the AzureCliTokenSource class. Once these are addressed, LGTM.

databricks/sdk/core.py
```diff
@@ -253,32 +253,69 @@ def refresh(self) -> Token:
 class AzureCliTokenSource(CliTokenSource):
     """ Obtain the token granted by `az login` CLI command """
 
-    def __init__(self, resource: str):
+    def __init__(self, resource: str, subscription: str = ""):
         cmd = ["az", "account", "get-access-token", "--resource", resource, "--output", "json"]
+        if subscription != "":
+            cmd.append("--subscription")
+            cmd.append(subscription)
```
tanmay-db, line 260: nit: we can use `extend` if we are adding more than one item. It would be more readable.

databricks/sdk/core.py
```diff
@@ -253,32 +253,69 @@ def refresh(self) -> Token:
...
 
+    @staticmethod
+    def for_resource(cfg: 'Config', resource: str) -> 'AzureCliTokenSource':
+        subscription = AzureCliTokenSource._get_subscription(cfg)
+        if subscription != "":
+            token = AzureCliTokenSource(resource, subscription)
+            try:
+                # This will fail if the user has access to the workspace, but not to the subscription
+                # itself.
+                # In such case, we fall back to not using the subscription.
```
tanmay-db, line 274: nit: the bottom comment can be moved above.

databricks/sdk/core.py
```diff
@@ -253,32 +253,69 @@ def refresh(self) -> Token:
...
+            )
+        token = AzureCliTokenSource(resource)
+        token.token()
+        return token
+
+    @staticmethod
+    def _get_subscription(cfg: 'Config') -> str:
+        resource = cfg.azure_workspace_resource_id
+        if resource == None or resource == "":
+            return ""
```
tanmay-db, line 292: We may want to log something along the lines of resource not found.

tanmay-db (approved):
couple of nits otherwise LGTM.

PR #347, opened 2023-09-15
Add ErrorInfo to API errors

hectorcast-db:
## Changes

Databricks API can return error details in case of errors. In some cases, users need to access such details to be able to solve the issue. This is the case for the errors of type ErrorInfo used by the Settings Platform.
This PR adds the necessary fields to the APIError to Unmarshal the ErrorDetails and to expose the details of type ErrorInfo to the users.

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -925,7 +925,7 @@ class DatabricksError(IOError):
                  status: str = None,
                  scimType: str = None,
                  error: str = None,
-                 details: List[dict[str, any]] = [],
+                 details: List[Dict[str, any]] = [],
                  **kwargs):
         if error:
             # API 1.2 has different response format, let's adapt
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -925,7 +925,7 @@ class DatabricksError(IOError):
                  status: str = None,
                  scimType: str = None,
                  error: str = None,
-                 details: List[dict] = [],
+                 details: List[dict[str, any]] = [],
                  **kwargs):
         if error:
             # API 1.2 has different response format, let's adapt
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -925,7 +925,7 @@ class DatabricksError(IOError):
                  status: str = None,
                  scimType: str = None,
                  error: str = None,
-                 details: List[ErrorDetail] = [],
+                 details: List[dict] = [],
                  **kwargs):
         if error:
             # API 1.2 has different response format, let's adapt
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -892,6 +892,30 @@ class Config:
         return cpy
 
 
+# Known ErrorDetail types
+errorInfoType = "type.googleapis.com/google.rpc.ErrorInfo"
+
+
+class ErrorDetail:
+
+    def __init__(self,
+                 type: str = None,
+                 reason: str = None,
+                 domain: str = None,
+                 metadata: dict = None,
+                 **kwargs):
+        self.type = type
+        self.reason = reason
+        self.domain = domain
+        self.metadata = metadata
+
+    @classmethod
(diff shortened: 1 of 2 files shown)


databricks/sdk/core.py
```diff
@@ -919,8 +944,17 @@ def __init__(self,
             error_code = f"SCIM_{status}"
         super().__init__(message if message else error)
         self.error_code = error_code
+        self.details = [ErrorDetail.from_dict(detail) for detail in details]
         self.kwargs = kwargs
 
+    def GetErrorInfo(self) -> List[ErrorDetail]:
```
pietern, line 950: Function and variable names must be snake-cased here.

databricks/sdk/core.py
```diff
@@ -919,8 +944,17 @@ def __init__(self,
...
+        return self.getDetailsByType(errorInfoType)
+
+    def getDetailsByType(self, errorType) -> List[ErrorDetail]:
```
pietern, line 953: Same.

databricks/sdk/core.py
```diff
@@ -892,6 +892,30 @@ def copy(self):
         return cpy
 
 
+# Known ErrorDetail types
+errorInfoType = "type.googleapis.com/google.rpc.ErrorInfo"
+
```
pietern, line 897: This should move into the DatabricksError class and be prefixed with a `_` to indicate it's not a part of the public API.

Also snake-case please.

databricks/sdk/core.py
```diff
@@ -919,8 +944,17 @@ def __init__(self,
             error_code = f"SCIM_{status}"
         super().__init__(message if message else error)
         self.error_code = error_code
+        self.details = [ErrorDetail.from_dict(detail) for detail in details]
         self.kwargs = kwargs
 
+    def get_error_info(self) -> List[ErrorDetail]:
+        return self._get_details_by_type(errorInfoType)
+
+    def _get_details_by_type(self, errorType) -> List[ErrorDetail]:
```
pietern, line 953: Please also snake-case.

tests/test_core.py
```diff
@@ -285,3 +286,38 @@ def test_api_client_do_custom_headers(config, requests_mock):
...
+    with pytest.raises(DatabricksError) as raised:
+        client.do("GET", "/test", headers={"test": "test"})
+
+    errorInfos = raised.value.get_error_info()
+    assert len(errorInfos) == 1
+    errorInfo = errorInfos[0]
+    assert errorInfo.reason == "errorReason"
+    assert errorInfo.domain == "errorDomain"
+    assert errorInfo.metadata["etag"] == "errorEtag"
+    assert errorInfo.type == errorInfoType
```
pietern, line 323: Please use snake-case everywhere here. I believe we only use PascalCase for class names in our Python code.

databricks/sdk/core.py
```diff
@@ -919,13 +915,16 @@ def from_dict(cls, d: Dict[str, any]) -> 'ErrorDetail':
...
+    _error_info_type = "type.googleapis.com/google.rpc.ErrorInfo"
+
     def __init__(self,
                  message: str = None,
                  *,
                  error_code: str = None,
                  detail: str = None,
                  status: str = None,
-                 scimType: str = None,
+                 scim_type: str = None,
```
pietern, line 927: We can't rename this field; this is what the field is named in the response body...

hectorcast-db: I noticed just after pushing. Added a test since this could be changed without noticing on a bigger PR.

databricks/sdk/core.py
```diff
@@ -903,6 +925,7 @@ def __init__(self,
                  status: str = None,
                  scimType: str = None,
                  error: str = None,
+                 details: List[ErrorDetail] = [],
```
pietern, line 928: On second look, this type hint is incorrect because the constructor expects a `Dict[str, any]`.

pietern (approved):
Everything else LGTM.

PR #361, opened 2023-09-22
Use `[]` instead of `None` as default list value for deserialising responses

nfx:
## Changes

This PR makes the experience consistent with Go SDK for the read path. Currently, we use `None` as the default value for list fields. This may result in unexpected runtime bugs like

```
_backup = w.groups.get(...)
_backup_members = sorted([m.value for m in _backup.members])
...

E           TypeError: 'NoneType' object is not iterable
```

Technically, we can ask users to be very defensive in programming and do `sorted(m.value for m in _backup.members) if _backup.members else []`, but most of the intended users do not know/want to know about these details.

What other pros and cons does keeping `None` as a default value?

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied

### DIFF
--- a/databricks/sdk/service/_internal.py
+++ b/databricks/sdk/service/_internal.py
@@ -10,7 +10,7 @@ def _from_dict(d: Dict[str, any], field: str, cls: Type) -> any:
 
 def _repeated(d: Dict[str, any], field: str, cls: Type) -> any:
     if field not in d or not d[field]:
-        return None
+        return []
     from_dict = getattr(cls, 'from_dict')
     return [from_dict(v) for v in d[field]]
 


alexott:
One thing to consider - do we need to distinguish in responses a "no result" vs. an "empty result"? If we have something as an array or dict, then an empty result could be a good default, and decrease a number of errors.

pietern:
If applied, we'd need to check for the same in Java and JavaScript.

In Go, users can still differentiate between the `None` and `[]` equivalents. It's just that iterating over `nil` is valid.

databricks/sdk/service/_internal.py
```diff
@@ -10,7 +10,7 @@ def _from_dict(d: Dict[str, any], field: str, cls: Type) -> any:
 
 def _repeated(d: Dict[str, any], field: str, cls: Type) -> any:
     if field not in d or not d[field]:
-        return None
+        return []
```
lennartkats-db, line 13: The main thing I'd review is how Databricks proto deserialization handles empty lists and if there are cases where APIs POST APIs don't support an empty array. For example, the `tasks` field in jobs comes to mind. In the case of Jobs it's fine to pass [] — we don't actually allow jobs with zero tasks but ignore this array if there is a task in the settings body of the job. Are there other cases where `[]` would break

nfx:
@pietern 

> In Go, users can still differentiate between the None and [] equivalents. It's just that iterating over nil is valid.

I'm afraid you are not correct; `[]string` zeroval is `[]string{}` (empty slice). https://go.dev/play/p/12J2n0JIC7_z

nfx:
> distinguish in responses a "no result" vs. an "empty result"

@alexott , what is the case for responses?

databricks/sdk/service/_internal.py
nfx: @lennartkats-db this is about deserialization in the SDK layer - of what Databricks platform has returned. it has nothing to do with how SDK requests are serialized.

specifically, this makes response deserialization to behave more similar to how proto deserializes

pietern:
@nfx 

The zeroval of a slice is nil, see: https://go.dev/play/p/3dGHbkXOJ_R

alexott:
My main concern is that we may have _some APIs_ (not sure if we have them, but requires investigation) that distinguish between not set and empty array, and then when we do the following chain of calls:

1. Set something without a field
2. Read it back -> "not set" will be converted into "empty array"
3. Modify the result & set it again -> APIs may complain about that.

@nkvuong do you remember anything similar in any API?

nfx:
@pietern but the point is that it's iterable - https://go.dev/play/p/12J2n0JIC7_z

nfx:
> Set something without a field
> Read it back -> "not set" will be converted into "empty array"

that's correct, that's what this PR is about.

> Modify the result & set it again -> APIs may complain about that.

which cases?

alexott:
I can't say it out of my head - maybe we can do a quick test across some existing assets that could confirm that?

databricks/sdk/service/_internal.py
lennartkats-db: From what you told me this influences if you get a job without a task property back from the SDK. Or a job with an empty array as the tasks property.

The challenge is that if we return an empty array, then it needs to be possible to submit that same JSON blob back to the API with e.g. the jobs create API. Otherwise users get very surprising behavior.

mgyucht: To clarify: this isn't about the JSON blob, it's about the materialized job object that is returned. The equivalent of what you're asking about in the SDK is whether someone can do
```py
j = w.jobs.get(job_id=123)
w.jobs.create(**j.settings)
```
In my understanding, the protobuf library does not distinguish between missing and empty list for repeated fields. If an API needs to be able to distinguish between these cases (e.g. an Update API where one of the fields is a list that can be optionally updated), a field mask or other mechanism is necessary to explicitly inform the server whether this is needed or not.

lennartkats-db: @mgyucht 
> j = w.jobs.get(job_id=123)
> w.jobs.create(**j.settings)

👍 We're talking about the same thing here.

So I mainly just want confirmation on
> The main thing I'd review is how Databricks proto deserialization handles empty lists and if there are cases where APIs POST APIs don't support an empty array.

It sounds like in your understanding, proto deserialization addresses this issue. I suspect that you're right. But could you 100% confirm that understanding?

mgyucht: I'll double-check with @alexc-db, he probably would be the best to confirm this.

nfx:
image
this is really causing flaky tests, as Jobs API and SCIM API are not sending the right responses.

mgyucht (approved):
This should be safe specifically for lists. Protobuf backends have no way to distinguish between an empty list or absence of a field (other types would be able to).

PR #373, opened 2023-10-02
Update OpenAPI spec to 2 Oct 2023

mgyucht:
## Changes
This PR updates the OpenAPI spec for Databricks SDK for Python. The main noteworthy changes are:
1. Changing the deserialization pathway for paginated, non-multi-request APIs (i.e., APIs that return an object containing an array of all listed values). Now, we will use the `from_dict` method for the Response class and return the named result field. "Multi-request" APIs are not changed (they still use a generator and fetch new pages incrementally on demand), and list APIs that return an array directly are not handled by the `method-call-paginated` template.

## Tests
Reran all integration tests.
Manually ran some code exercising the updated pathways, such as:
```
a.metastore_assignments.list(y.metastore_assignment.metastore_id)
```


.codegen/service.py.tmpl
```diff
@@ -275,7 +275,7 @@ class {{.Name}}API:{{if .Description}}
           {{- end}}
         {{else -}}
         json = {{template "method-do" .}}
-        return [{{.Pagination.Entity.PascalName}}.from_dict(v) for v in json{{if .Pagination.Results}}.get('{{.Pagination.Results.Name}}', []){{end}}]
+        return {{.Response.PascalName}}.from_dict(json).{{.Pagination.Results.Name}}
```
nfx, line 278: this change will cause downstream code to break if the response does not contain `.Pagination.Results.Name` value in it, because `None` is not iterable.

nfx (changes requested):

.codegen/service.py.tmpl
```diff
@@ -312,7 +312,7 @@ self._api.do('{{.Verb}}',
   {{- else if .Response}} -> {{if .Response.ArrayValue -}}
     Iterator[{{template "type" .Response.ArrayValue}}]
   {{- else if .Pagination -}}
-    Iterator[{{.Pagination.Entity.PascalName}}]
+    Iterator[{{template "type" .Pagination.Entity}}]
```
nfx, line 315: do we iterate over `str` and `int` somewhere?

mgyucht: Yes: list metastore assignments response is a list of workspace IDs, which are plain ints.

nfx (approved):

tanmay-db (approved):

PR #380, opened 2023-10-04
Support workload_type field for ServedModel in SDK

ankit-db:
## Changes
We've added `workload_type` as a field to the ServedModel spec - our customers have expressed a desire to use the SDK to control the value of this parameter and create GPU endpoints. This feature is now in public preview, so we are documenting this.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


nfx:
@ankit-db this has to be a change to OpenAPI spec, not SDK. add it in universe

PR #385, opened 2023-10-04
Add support for tokenless authentication for GitHub Actions configured with OpenID Connect with Azure User Managed Identity (or Service Principal)

nfx:
## Changes
See https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-cloud-providers and https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-azure

Technically, it should also work with Azure DevOps Workload Identity Federation, once we figure out the environment variables: https://techcommunity.microsoft.com/t5/azure-devops-blog/introduction-to-azure-devops-workload-identity-federation-oidc/ba-p/3908687

## Tests

setup:
```
resource "github_actions_environment_secret" "scope" {
  for_each    = github_repository_environment.default
  repository  = each.value.repository
  environment = each.value.environment
  secret_name = "ARM_CLIENT_ID"
  # this value is not a secret as well
  plaintext_value = data.azurerm_user_assigned_identity.scopes[local.project_scopes[each.key]].client_id
}

...

resource "azurerm_federated_identity_credential" "oidc" {
  for_each            = github_repository_environment.default
  name                = "${local.org}-${each.value.repository}-${each.value.environment}-oidc"
  resource_group_name = local.resource_group_name
  audience            = ["api://AzureADTokenExchange"]
  issuer              = "https://token.actions.githubusercontent.com"
  parent_id           = data.azurerm_user_assigned_identity.scopes[local.project_scopes[each.key]].id
  subject             = "repo:${local.org}/${each.value.repository}:environment:${each.value.environment}"
}

...

resource "azurerm_user_assigned_identity" "scopes" {
  for_each = local.scopes
  name     = "labs-${each.key}-identity"

  resource_group_name = azurerm_resource_group.this.name
  location            = azurerm_resource_group.this.location
  // ...
}

...

resource "databricks_service_principal" "scopes" {
  provider = databricks.account
  for_each = local.scopes

  application_id = azurerm_user_assigned_identity.scopes[each.key].client_id
  display_name   = azurerm_user_assigned_identity.scopes[each.key].name
  external_id    = azurerm_user_assigned_identity.scopes[each.key].principal_id
}
```

result

_Experiment__Call_integration_tests_via_OIDC_·_databrickslabs_ucx_5a94b24

- [ ] `make test` run locally
- [x] `make fmt` applied

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -15,16 +15,17 @@ import urllib.parse
 from datetime import datetime, timedelta
 from json import JSONDecodeError
 from types import TracebackType
-from typing import (Any, BinaryIO, Callable, Dict, Iterable, Iterator, List, Optional, Type, Union)
+from typing import (Any, BinaryIO, Callable, Dict, Iterable, Iterator, List,
+                    Optional, Type, Union)
 
 import requests
 import requests.auth
 from requests.adapters import HTTPAdapter
 
-from .azure import (ARM_DATABRICKS_RESOURCE_ID, ENVIRONMENTS, AzureEnvironment, add_sp_management_token,
-                    add_workspace_id_header)
-from .oauth import (ClientCredentials, OAuthClient, OidcEndpoints, Refreshable, Token, TokenCache,
-                    TokenSource)
+from .azure import (ARM_DATABRICKS_RESOURCE_ID, ENVIRONMENTS, AzureEnvironment,
+                    add_sp_management_token, add_workspace_id_header)
+from .oauth import (ClientCredentials, OAuthClient, OidcEndpoints, Refreshable,
+                    Token, TokenCache, TokenSource)
 from .retries import retried
 from .version import __version__
 
@@ -99,7 +100,8 @@ def runtime_native_auth(cfg: 'Config') -> Optional[HeaderFactory]:
     # This import MUST be after the "DATABRICKS_RUNTIME_VERSION" check
     # above, so that we are not throwing import errors when not in
     # runtime and no config variables are set.
-    from databricks.sdk.runtime import (init_runtime_legacy_auth, init_runtime_native_auth,
+    from databricks.sdk.runtime import (init_runtime_legacy_auth,
+                                        init_runtime_native_auth,
                                         init_runtime_repl_auth)
     for init in [init_runtime_native_auth, init_runtime_repl_auth, init_runtime_legacy_auth]:
         if init is None:
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -30,6 +30,7 @@ from .retries import retried
 from .version import __version__
 
 __all__ = ['Config', 'DatabricksError']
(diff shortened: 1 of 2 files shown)


databricks/sdk/core.py
```diff
@@ -217,6 +217,47 @@ def refreshed_headers() -> Dict[str, str]:
...
+    if not cfg.is_azure:
+        return None
+
+    # See https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-cloud-providers
+    headers = {'Authorization': f"Bearer {os.environ['ACTIONS_ID_TOKEN_REQUEST_TOKEN']}"}
+    endpoint = f"{os.environ['ACTIONS_ID_TOKEN_REQUEST_URL']}&audience=api://AzureADTokenExchange"
+    response = requests.get(endpoint, headers=headers)
+
+    # get the ID Token with aud=api://AzureADTokenExchange sub=repo:org/repo:environment:name
+    client_assertion = response.json()['value']
```
mgyucht, line 237: If the request fails, this throws an exception. I think we need to return `None` instead and log a warning.

databricks/sdk/core.py
```diff
@@ -217,6 +217,47 @@ def refreshed_headers() -> Dict[str, str]:
...
+        'client_assertion_type': 'urn:ietf:params:oauth:client-assertion-type:jwt-bearer',
+        'resource': cfg.effective_azure_login_app_id,
+        'client_assertion': client_assertion,
+    }
+    aad_endpoint = cfg.arm_environment.active_directory_endpoint
+    if not cfg.azure_tenant_id:
+        # detect Azure AD Tenant ID if it's not specified directly
+        cfg.azure_tenant_id = cfg.oidc_endpoints.token_endpoint.replace(aad_endpoint, '').split('/')[0]
+    inner = ClientCredentials(client_id=cfg.azure_client_id,
+                              client_secret=cfg.azure_client_secret,
```
mgyucht, line 250: Is this used in this workflow? I would figure this should be `None` or `""` for the OIDC flow since the OIDC token is passed in params.

mgyucht (commented):
Seems very reasonable, thanks for contributing. Added a new row to our unified auth template for SDKs to support this. Over time, we may want to generalize to support e.g. M2M SP via OIDC if Auth Platform is interested in supporting it; I asked a question on their channel. That said, I think this is a reasonable way to get started and help users who want a secret-free experience.

Can you add a couple of unit tests, esp. checking the case where the OIDC token endpoint times out or doesn't respond with a 200/JSON?

databricks/sdk/core.py
nfx: yep, fixed

mgyucht (commented):
Implementation seems fine to me, but let's not include test code paths in our main code. I think you just need to specify an azure resource ID for `is_azure()` to return true.

nfx:
@mgyucht it's not possible to do full unit testing without a special test path.

mgyucht (approved):

PR #388, opened 2023-10-05
Fix select spark version in staging

mgyucht:
## Changes
In staging, there are some spark versions that do not have a patch version. We fall back to a different pattern with no patch component for these. Additionally, we check if the builds are equal to one another (which allows us to avoid comparing one < the other in the case where they are both None).

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/databricks/sdk/mixins/compute.py
+++ b/databricks/sdk/mixins/compute.py
@@ -27,6 +27,11 @@ class SemVer:
                           r"(?:-(?P<pre_release>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)"
                           r"(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?"
                           r"(?:\+(?P<build>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$")
+    _pattern_no_patch = re.compile(r"^"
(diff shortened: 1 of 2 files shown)


databricks/sdk/mixins/compute.py
```diff
@@ -21,9 +21,9 @@ class SemVer:
     build: Optional[str] = None
 
     # official https://semver.org/ recommendation: https://regex101.com/r/Ly7O1x/
-    # with addition of "x" wildcards for minor/patch versions
+    # with addition of "x" wildcards for minor/patch versions. Also, patch version may be omitted.
     _pattern = re.compile(r"^"
-                          r"(?P<major>0|[1-9]\d*)\.(?P<minor>x|0|[1-9]\d*)\.(?P<patch>x|0|[1-9x]\d*)"
+                          r"(?P<major>0|[1-9]\d*)\.(?P<minor>x|0|[1-9]\d*)(\.(?P<patch>x|0|[1-9x]\d*))?"
```
tanmay-db, line 26: I understand we make patch version optional so it works for snapshot (staging) but won't this make the semver less strict for production use cases?

tanmay-db (approved):

databricks/sdk/mixins/compute.py
nfx: we can always check on prod :)

mgyucht: Do you think there is a risk there?

mgyucht: I tried this in prod, and it worked there as well.

PR #390, opened 2023-10-06
Added `w.config.account_host` to get the relevant account host from a workspace client

nfx:
## Changes
This simplifies downstream integration testing as well as configuration

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


databricks/sdk/azure.py
```diff
@@ -16,18 +17,23 @@ class AzureEnvironment:
...
                             service_management_endpoint="https://management.core.windows.net/",
                             resource_manager_endpoint="https://management.azure.com/",
                             active_directory_endpoint="https://login.microsoftonline.com/"),
-    GERMAN=AzureEnvironment(name="AzureGermanCloud",
-                            service_management_endpoint="https://management.core.cloudapi.de/",
-                            resource_manager_endpoint="https://management.microsoftazure.de/",
-                            active_directory_endpoint="https://login.microsoftonline.de/"),
+    GERMAN=AzureEnvironment(
+        name="AzureGermanCloud",
+        databricks_hostname_suffix='.databricks.azure.de', # TODO: i'm not sure about this one
```
mgyucht, line 26: I think this is OK because we don't run Databricks in this cloud (and it is being deprecated: https://learn.microsoft.com/en-us/previous-versions/azure/germany/germany-welcome)

databricks/sdk/core.py
```diff
@@ -617,8 +613,31 @@ def is_account_client(self) -> bool:
             return False
         return self.host.startswith("https://accounts.") or self.host.startswith("https://accounts-dod.")
 
+    @property
+    def account_host(self) -> str:
```
mgyucht, line 617: This won't work for dev/staging, right? We occasionally get internal issues where the SDK doesn't work well with dev/staging.

Also, can we label this somehow as experimental? In the future, if we introduce a workspace-level metadata service, the accounts endpoint can be exposed there, and we can remove this property.

mgyucht:
One nit to make this work for dev/staging & label as experimental, but otherwise this seems reasonable to me if it unblocks you/customers.

databricks/sdk/core.py
nfx: can you suggest on the pattern here to make it work for dev/staging?

btw, once we introduce workspace-level metadata service, we should just print warnings from within this call.

mgyucht: One idea: have a list to lookup cloud/env/account host name based on domain suffix, from most-specific to least-specific:
```
    # Most specific to least specific
    [
        ('.dev.databricks.com', 'AWS', 'DEV', 'PUBLIC', 'accounts.dev.databricks.com'),
        ('.staging.cloud.databricks.com', 'AWS', 'STAGING', 'PUBLIC', 'accounts.staging.cloud.databricks.com'),
        ('.cloud.databricks.com', 'AWS', 'PROD', 'PUBLIC', 'accounts.cloud.databricks.com'),
        ('.cloud.databricks.us', 'AWS', 'PROD', 'GOV', 'accounts.cloud.databricks.us'),
        ('.dev.azuredatabricks.net', 'AZURE', 'DEV', 'PUBLIC', 'accounts.dev.azuredatabricks.net')
        ('.staging.azuredatabricks.net', 'AZURE', 'STAGING', 'PUBLIC', 'accounts.staging.azuredatabricks.net'),
        ('.azuredatabricks.net', 'AZURE', 'PROD', 'PUBLIC', 'accounts.azuredatabricks.net'),
        ('.databricks.azure.us', 'AZURE', 'PROD', 'GOV', 'accounts.databricks.azure.us'),
        ('.dev.gcp.databricks.com', 'GCP', 'DEV', 'PUBLIC', 'accounts.dev.gcp.databricks.com'),
        ('.staging.gcp.databricks.com', 'GCP', 'STAGING', 'PUBLIC', 'accounts.staging.gcp.databricks.com'),
        ('.gcp.databricks.com', 'GCP', 'PROD', 'PUBLIC', 'accounts.gcp.databricks.com'),
    ]
```
The simpler version is to just replace the most specific zone with `accounts`, which works everywhere today.

nfx: @mgyucht do we want to expose dev/staging?... if we do - then why don't we expose the `azure_login_app_id` along with it?

hamza-db:
all the info here is fine to be public ( and already public)

databricks/sdk/core.py
```diff
@@ -506,6 +508,56 @@ def __repr__(self) -> str:
...
+    tld: str
+
+    azure_environment: AzureEnvironment = None
+
+    # The application (client) ID isn't a secret. See
+    # https://learn.microsoft.com/en-us/entra/identity-platform/developer-glossary#application-client-id
+    azure_application_id: str = None
+
+    def deployment(self, deployment_name: str) -> str:
+        return f'https://{deployment_name}{self.tld}'
```
mgyucht, line 531: This is maybe just a bit abstract. It won't work in Azure the way a user might expect, where the DNS name can't be determined from the workspace ID alone. Given we're only using this for account host, I suggest we just inline it there.

mgyucht (approved):
This is great, thank you for contributing! Couple nits/suggestions, otherwise this looks good to me.

databricks/sdk/core.py
nfx: well... in this case we have to specify `adb-633....10` as deployment name

https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/account/workspaces.py#L101-L126

databricks/sdk/core.py
```diff
@@ -506,6 +508,56 @@ def __repr__(self) -> str:
...
+    AZURE = 'azure'
+    GCP = 'gcp'
+
+
+@dataclass
+class DatabricksEnvironment:
+    cloud: Cloud
+
+    # domain suffixes are not very secret: https://crt.sh/?q=databricks
+    tld: str
```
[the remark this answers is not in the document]
nfx: `dns_zone` it is then.

databricks/sdk/core.py
mgyucht: Oh perfect, I actually didn't realize those workspace APIs worked in azure yet.

nfx:
this needs a bit more adjustments to work

pietern:
Are similar PRs for the other SDKs tracked some place?

nfx:
@pietern i'll have one for GoLang the following week, need the same thing for two projects

renaudhartert-db:

PR #395, opened 2023-10-10
Add retries on `RPC token bucket limit has been exceeded`

nfx:
## Changes

Another case when the platform is not sending a proper HTTP 429 response. This time for the SCIM user list:
image

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


mgyucht (approved):

pietern:
@nfx @mgyucht Do we have a canonical format for these, or is this just a message substring?

It would be nice to replicate these conditions across SDKs, as we do for the auth permutations.

mgyucht:
@pietern I'm filing a follow-up with the platform team to fix the underlying issue.

PR #406, opened 2023-10-13
[JOBS-11439] Add taskValues support in remoteDbUtils

vsamoilov:
## Changes
Support `dbutils.jobs.taskValues` methods in `RemoteDbUtils` and add additional `checkTaskValue` to allow local testing

## Tests

Added three test cases in `test_dbutils.py`. Integration tests are not needed since changes only for remote mode
- [x] `make test` run locally
- [x] `make fmt` applied

### DIFF
--- a/databricks/sdk/dbutils.py
+++ b/databricks/sdk/dbutils.py
@@ -174,7 +174,9 @@ class _JobsUtil:
             Returns `debugValue` if present, throws an error otherwise as this implementation is always run outside of a job run
             """
             if debugValue is None:
-                raise ValueError('No debug value set for task value, when outside of job run')
+                raise TypeError(
+                    'Must pass debugValue when calling get outside of a job context. debugValue cannot be None.'
+                )
             return debugValue
 
         def set(self, key: str, value: any) -> None:
--- a/databricks/sdk/dbutils.py
+++ b/databricks/sdk/dbutils.py
@@ -171,8 +171,10 @@ class _JobsUtil:
 
         def get(self, taskKey: str, key: str, default: any = None, debugValue: any = None) -> None:
             """
-            Returns the latest task value that belongs to the current job run
+            Returns `debugValue` if present, throws an error otherwise as this implementation is always run outside of a job run
             """
+            if debugValue is None:
+                raise ValueError('No debug value set for task value, when outside of job run')
             return debugValue
 
         def set(self, key: str, value: any) -> None:
--- a/databricks/sdk/dbutils.py
+++ b/databricks/sdk/dbutils.py
@@ -163,6 +163,35 @@ class _SecretsUtil:
         return [SecretScope(v.name) for v in self._api.list_scopes()]
 
 
+class _JobsUtil:
+    """Remote equivalent of jobs util"""
+
+    class _TaskValuesUtil:
+        """Remote equivalent of task values util"""
+        values = {}
+
+        def get(self, taskKey: str, key: str, default: any = None, debugValue: any = None) -> None:
+            """
+            Returns the latest task value that belongs to the current job run
+            """
+            return debugValue
+
+        def set(self, key: str, value: any) -> None:
+            """
+            Sets a task value on the current task run
+            """
+            self.values[key] = value
+
+        def checkTaskValue(self, key: str, value: any) -> bool:
+            """
+            Check that the task value with given key is equal to the given value
+            """
+            return self.values.get(key) == value
+
+    def __init__(self) -> None:
+        self.taskValues = self._TaskValuesUtil()
+
+
 class RemoteDbUtils:
 
     def __init__(self, config: 'Config' = None):
@@ -175,6 +204,7 @@ class RemoteDbUtils:
 
         self.fs = _FsUtil(dbfs_ext.DbfsExt(self._client), self.__getattr__)
         self.secrets = _SecretsUtil(workspace.SecretsAPI(self._client))
+        self.jobs = _JobsUtil()
         self._widgets = None
 
     # When we import widget_impl, the init file checks whether user has the
--- a/databricks/sdk/dbutils.py
+++ b/databricks/sdk/dbutils.py
@@ -168,7 +168,6 @@ class _JobsUtil:
 
     class _TaskValuesUtil:
         """Remote equivalent of task values util"""
-        values = {}
 
         def get(self, taskKey: str, key: str, default: any = None, debugValue: any = None) -> None:
             """
@@ -180,13 +179,6 @@ class _JobsUtil:
             """
             Sets a task value on the current task run
             """
-            self.values[key] = value
-
-        def checkTaskValue(self, key: str, value: any) -> bool:
-            """
-            Check that the task value with given key is equal to the given value
-            """
-            return self.values.get(key) == value
 
     def __init__(self) -> None:
         self.taskValues = self._TaskValuesUtil()
--- a/tests/test_dbutils.py
+++ b/tests/test_dbutils.py
@@ -204,3 +204,21 @@ def test_secrets_get_and_redacting_logs(dbutils, mocker):
     inner.assert_called_with('GET', '/api/2.0/secrets/get', query={'key': 'bar', 'scope': 'foo'})
 
     assert value == 'hello'
+
+
+def test_jobs_task_values_check_task_value(dbutils):
+    assert (dbutils.jobs.taskValues.checkTaskValue('key', 'value') == False)
+
+
+def test_jobs_task_values_set(dbutils):
+    dbutils.jobs.taskValues.set('key', 'value')
+
+    assert dbutils.jobs.taskValues.checkTaskValue('key', 'value')
+
+
+def test_jobs_task_values_get(dbutils):
+    assert dbutils.jobs.taskValues.get('taskKey', 'key', debugValue='debug') == 'debug'
+
+    dbutils.jobs.taskValues.set('key', 'value')
+
+    assert dbutils.jobs.taskValues.get('taskKey', 'key', debugValue='debug') == 'debug'
--- a/tests/test_dbutils.py
+++ b/tests/test_dbutils.py
@@ -215,6 +215,7 @@ def test_jobs_task_values_get(dbutils):
 
     dbutils.jobs.taskValues.set('key', 'value')
 
+    # Expect `get` to always return the `debugValue`` when calling outside of a job context and not what was previously set using `set`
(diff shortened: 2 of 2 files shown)


databricks/sdk/dbutils.py
```diff
@@ -163,6 +163,35 @@ def listScopes(self) -> typing.List[SecretScope]:
...
+            """
+            Sets a task value on the current task run
+            """
+            self.values[key] = value
+
+        def checkTaskValue(self, key: str, value: any) -> bool:
+            """
+            Check that the task value with given key is equal to the given value
+            """
+            return self.values.get(key) == value
```
mgyucht, line 189: We should make sure _TaskValuesUtil implements the interface defined at https://github.com/databricks/databricks-sdk-py/blob/main/databricks/sdk/runtime/dbutils_stub.py#L197. This only defines get and set. Otherwise, people may use this method and be surprised that it doesn't exist in DBR.

databricks/sdk/dbutils.py
```diff
@@ -163,6 +163,35 @@ def listScopes(self) -> typing.List[SecretScope]:
...
+
+    class _TaskValuesUtil:
+        """Remote equivalent of task values util"""
+        values = {}
+
+        def get(self, taskKey: str, key: str, default: any = None, debugValue: any = None) -> None:
+            """
+            Returns the latest task value that belongs to the current job run
+            """
+            return debugValue
```
mgyucht, line 177: Wow, this API is a bit confusing. What is the difference between `default` and `debugValue`?

vsamoilov: https://docs.databricks.com/en/dev-tools/databricks-utils.html#get-command-dbutilsjobstaskvaluesget - `default` is returned when we inside job run and task value can't be found. And `debugValue` is what we always return when outside of job: for example, to be able to run notebook outside of job and check if it is just working without errors

vsamoilov:
> I'm having trouble really understanding this change. Seems very simple, but maybe I'm confused about the API we are trying to support. Is it not possible to make this work for testing without exposing a new method? I would have thought that you could set a task value and then get it; I think you can do that in DBR but you can't outside, but we can't mock that behavior? Do people ever specify different values for default and debugValue?

I was implementing change according to that doc: https://docs.google.com/document/d/1yrXXA27GJOB968apgnxnMsWfj0SJJn2cpArE4zzIORU/edit#heading=h.a2at0pdkk01r. As I understood the idea for additional method is to check if `set(key, value)` was called with specified parameters, because `get` interface is for case when we are inside job run (we need some task key at least) cc @gaborratky-db @mgyucht

databricks/sdk/dbutils.py
```diff
@@ -163,6 +163,35 @@ def listScopes(self) -> typing.List[SecretScope]:
...
+class _JobsUtil:
+    """Remote equivalent of jobs util"""
+
+    class _TaskValuesUtil:
+        """Remote equivalent of task values util"""
+        values = {}
+
+        def get(self, taskKey: str, key: str, default: any = None, debugValue: any = None) -> None:
+            """
+            Returns the latest task value that belongs to the current job run
```
gaborratky-db, line 175:
```suggestion
            Returns `debugValue` if present, throws an error otherwise as this implementation is always run outside of a job run
```

databricks/sdk/dbutils.py
gaborratky-db, line 189:
```suggestion
        def assert(self, key: str, value: any) -> None:
            """
            Assert that the task value with the given key was set to the given value. This method is only available when run outside of a job run (i.e., locally or as part of a unit test).
            """
            assert self.values.get(key) == value
```

tests/test_dbutils.py
```diff
@@ -204,3 +204,21 @@ def test_secrets_get_and_redacting_logs(dbutils, mocker):
...
+
+
+def test_jobs_task_values_check_task_value(dbutils):
+    assert (dbutils.jobs.taskValues.checkTaskValue('key', 'value') == False)
+
+
+def test_jobs_task_values_set(dbutils):
+    dbutils.jobs.taskValues.set('key', 'value')
+
+    assert dbutils.jobs.taskValues.checkTaskValue('key', 'value')
```
gaborratky-db, line 216:
```suggestion
    dbutils.jobs.taskValues.assert('key', 'value')
```

tests/test_dbutils.py
```diff
@@ -204,3 +204,21 @@ def test_secrets_get_and_redacting_logs(dbutils, mocker):
     inner.assert_called_with('GET', '/api/2.0/secrets/get', query={'key': 'bar', 'scope': 'foo'})
 
     assert value == 'hello'
+
+
+def test_jobs_task_values_check_task_value(dbutils):
+    assert (dbutils.jobs.taskValues.checkTaskValue('key', 'value') == False)
```
gaborratky-db, line 210:
```suggestion
    try:
        dbutils.jobs.taskValues.assert('key', 'value')
        // do something idiomatic here that makes the test fail
    except AssertError:
       // Swallow error as it's what we expected to happen
```

gaborratky-db (commented):
Here are my suggested changes based on my original thinking. If `assert` is confusing/not useful, then let's remove it.

databricks/sdk/dbutils.py
vsamoilov: I can't call it `assert`, is `assertTaskValue` works?

mgyucht:
Thanks for clarifying, that makes a lot of sense. My main concern is really that there is a divergence in the API between DBUtils in the SDK and DBUtils in DBR. Where possible, I think preventing this divergence ensures that customers don't run into AttributeErrors at runtime in DBR. I do understand why you need to add an additional method now because of the specific semantics of getting task values outside of a job run. If we add this method here, I suggest adding a corresponding method in DBR as well. Do you think that's feasible?

vsamoilov:
> Thanks for clarifying, that makes a lot of sense. My main concern is really that there is a divergence in the API between DBUtils in the SDK and DBUtils in DBR. Where possible, I think preventing this divergence ensures that customers don't run into AttributeErrors at runtime in DBR. I do understand why you need to add an additional method now because of the specific semantics of getting task values outside of a job run. If we add this method here, I suggest adding a corresponding method in DBR as well. Do you think that's feasible?

I think it is not very feasible, because if we want the similar implementation, then we need to check all job tasks for `taskValue` with specified `key`.  Also as I understood, use case during local development is that we can write unit tests in which we will run task function and then check that we set correct value in specified `key`. I don't know about such use cases in DBR. I think if we will write in documentation that it is only for unit test purposes during local development and implement it by `assert` call, then it should be enough to prevent using in DBR, what do you think? cc @mgyucht @gaborratky-db

gaborratky-db:
> > Thanks for clarifying, that makes a lot of sense. My main concern is really that there is a divergence in the API between DBUtils in the SDK and DBUtils in DBR. Where possible, I think preventing this divergence ensures that customers don't run into AttributeErrors at runtime in DBR. I do understand why you need to add an additional method now because of the specific semantics of getting task values outside of a job run. If we add this method here, I suggest adding a corresponding method in DBR as well. Do you think that's feasible?
> 
> I think it is not very feasible, because if we want the similar implementation, then we need to check all job tasks for `taskValue` with specified `key`. Also as I understood, use case during local development is that we can write unit tests in which we will run task function and then check that we set correct value in specified `key`. I don't know about such use cases in DBR. I think if we will write in documentation that it is only for unit test purposes during local development and implement it by `assert` call, then it should be enough to prevent using in DBR, what do you think? cc @mgyucht @gaborratky-db

This isn't necessarily true because we know in DBR which `taskKey` the `dbutils.jobs.taskValues` call runs for and we can call `dbutils.jobs.taskValues.get("<my own task key>", ... )` to retrieve the task value.

If we want to keep parity between DBR and the SDK implementation, one option is to extend the semantics of `dbutils.jobs.taskValues.get` and make it possible to query the task's own task values when called as part of a job run. For example:

`dbutils.jobs.taskValues.get(None, "myKey")` would query whether something was set with `dbutils.jobs.taskValues.set("myKey", "myValue")`. I don't like using `None` as a special `taskKey` value here but I don't have a better idea off the top of my head.

gaborratky-db:
I quickly looked at the implementation and we don't actually know the current `taskKey`, only the current run id, which we can't use to query the values.

I suggest that we simply remove `assertTaskValue` from the SDK version, keep parity, ship the fix, and come back to this problem when we have additional capacity to improve `dbutils.jobs.taskValues`.

databricks/sdk/dbutils.py
```diff
@@ -163,6 +163,29 @@ def listScopes(self) -> typing.List[SecretScope]:
...
+
+    class _TaskValuesUtil:
+        """Remote equivalent of task values util"""
+
+        def get(self, taskKey: str, key: str, default: any = None, debugValue: any = None) -> None:
+            """
+            Returns `debugValue` if present, throws an error otherwise as this implementation is always run outside of a job run
+            """
+            if debugValue is None:
+                raise ValueError('No debug value set for task value, when outside of job run')
```
gaborratky-db, line 177: Please throw the exact same error and message that DBR does:

```suggestion
                raise TypeError('Must pass debugValue when calling get outside of a job context. debugValue cannot be None.')
```

Based on:

image

tests/test_dbutils.py
```diff
@@ -204,3 +204,23 @@ def test_secrets_get_and_redacting_logs(dbutils, mocker):
...
+def test_jobs_task_values_set(dbutils):
+    dbutils.jobs.taskValues.set('key', 'value')
+
+
+def test_jobs_task_values_get(dbutils):
+    assert dbutils.jobs.taskValues.get('taskKey', 'key', debugValue='debug') == 'debug'
+
+    dbutils.jobs.taskValues.set('key', 'value')
+
+    assert dbutils.jobs.taskValues.get('taskKey', 'key', debugValue='debug') == 'debug'
```
gaborratky-db, line 218: Can you add a comment here that we expect `get` to always return the `debugValue` when calling outside of a job context and not what was previously set using `set`?

gaborratky-db (changes requested):
Two requests, LGTM otherwise. I will defer the final review to @mgyucht to see if he has any further questions.

gaborratky-db:
@mgyucht we addressed the changes and the implementation is simple and straightforward: the code will run in "remote" mode (aka locally) and will be more portable as a result. This should be enough for parity, PTAL.

mgyucht (approved):
I think this is good for now. If you are interested in updating the in-DBR dbutils.jobs.task_values to have other helpers in the future, we can make the corresponding changes here too.

PR #407, opened 2023-10-17
[BUG] fix spellcheck errror in `job.py` from `TIMEDOUT` to `TIMEOUT`

Future-Outlier:
# Changes
Currently, Flyte is integrating Databricks's service.
https://github.com/flyteorg/flytekit/pull/1797
We have to do a mapping mechanism to handle the response from Databricks API.
https://github.com/flyteorg/flytekit/blob/master/flytekit/extend/backend/base_agent.py#L138
But the spellcheck seems to be an error, please merge it, and thanks a lot!

# Test
- [ ] `make test` run locally
- [x] `make fmt` applied

### DIFF
--- a/databricks/sdk/service/jobs.py
+++ b/databricks/sdk/service/jobs.py
@@ -1853,7 +1853,7 @@ class RunParameters:
 
 class RunResultState(Enum):
     """A value indicating the run's result. The possible values are: * `SUCCESS`: The task completed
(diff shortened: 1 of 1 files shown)


nfx:
It is not allowed to change autogenerated files. It’s a valid enum state, please add exception to the spellcheck you’re using.

PR #408, opened 2023-10-17
Request management token via Azure CLI only for Service Principals and not human users

nfx:
## Changes
This PR removes the call to `az account get-access-token` to retrieve the management token if the machine is authenticated with a non-human user.

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -314,23 +314,42 @@ class AzureCliTokenSource(CliTokenSource):
                          access_token_field='accessToken',
                          expiry_field='expiresOn')
 
+    def is_human_user(self) -> bool:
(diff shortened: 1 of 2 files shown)


nfx:
Verified this by decoding claims from https://github.com/databricks/databricks-sdk-go/pull/659

hectorcast-db (approved):

mgyucht (commented):
Very nicely documented, thanks for this. Main comment is about the choice of field used to determine UPN vs SPN auth. For my understanding, is there a functional issue with the current implementation that this addresses, or is it just to reduce the unnecessary call to the Azure CLI?

PR #409, opened 2023-10-17
Remove noisy `azure_workspace_resource_id field not provided` message  for non-production Databricks deployments

nfx:
## Changes

JWT mismatched claims errors are only happening for dev and staging deployments of Databricks, adding the unnecessary noise to production users. `azure_workspace_resource_id` is not required for authentication to succeed.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied
- [ ] relevant integration tests applied

PR #410, opened 2023-10-18
Reduce redundant warning messages in azure-cli credential provider

mgyucht:
## Changes
This warning is a bit noisy: in the happy path, we print it multiple times, and even if we don't use Azure CLI auth, we may print it anyways. Also, it will be printed always when using account-level authentication.

This PR reduces the noise for this warning by only printing it when we are certainly going to use Azure CLI auth, authenticating to a workspace, and can't get the subscription ID (e.g. because it is not configured).

## Tests

- [x] Ran locally: with accounts client, no warning is printed, and with workspace client, only one warning is printed.

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -269,7 +269,7 @@ class AzureCliTokenSource(CliTokenSource):
 
     @staticmethod
     def for_resource(cfg: 'Config', resource: str) -> 'AzureCliTokenSource':
-        subscription = AzureCliTokenSource._get_subscription(cfg)
+        subscription = AzureCliTokenSource.get_subscription(cfg)
         if subscription != "":
             token = AzureCliTokenSource(resource, subscription)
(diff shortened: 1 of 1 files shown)


hectorcast-db (approved):

hectorcast-db:
Related PR (which will have a conflict with this)
https://github.com/databricks/databricks-sdk-py/pull/409/files

I was trying to remember which stage and credentials I used in my testing. Since you just tested it, can you confirm that this doesn't happen in production?

mgyucht:
@hectorcast-db I saw that PR which inspired me to do this one instead. After this change, this warning message will still be printed in production for users who don't specify the workspace resource ID. It can work if the workspace is in the same tenant as their current subscription, but for users who use multiple tenants, it can still be an issue, so I think the warning is well-placed to provide guidance for these otherwise inscrutable errors.

PR #419, opened 2023-11-01
Add more detailed error message on default credentials not found error

edwardfeng-db:
## Changes

- Added more detailed error message for default credentials not found error - found the original error message was a bit difficult to follow / understand

## Tests

Not quite sure if I need to do anything special if CI checks passed.
Manually tried using default auth locally without `.databrickscfg` file, verified that the prompt is printed out

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -527,7 +527,10 @@ class DefaultCredentials:
                 return header_factory
             except Exception as e:
                 raise ValueError(f'{auth_type}: {e}') from e
-        raise ValueError('cannot configure default credentials')
+        auth_flow_url = "https://docs.databricks.com/en/dev-tools/auth.html#databricks-client-unified-authentication"
+        raise ValueError(
+            f'cannot configure default credentials, please check {auth_flow_url} to configure credentials for your preferred authentication method.'
+        )
 
 
 class ConfigAttribute:
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -4,19 +4,24 @@ from databricks.sdk.core import Config
 
 from .conftest import __tests__, raises
 
+default_auth_base_error_message = \
+    "default auth: cannot configure default credentials, " \
+    "please check https://docs.databricks.com/en/dev-tools/auth.html#databricks-client-unified-authentication " \
+    "to configure credentials for your preferred authentication method"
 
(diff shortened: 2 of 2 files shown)


databricks/sdk/core.py
```diff
@@ -615,6 +615,10 @@ def wrap_debug_info(self, message: str) -> str:
         debug_string = self.debug_string()
         if debug_string:
             message = f'{message.rstrip(".")}. {debug_string}'
+        elif self._credentials_provider.auth_type() == "default":
+            # Empty debug string when auth type is default means we are likely missing some files or environment variables. Add a debug string to make it easier to understand.
+            default_auth_prompt = "Please check https://github.com/databricks/databricks-sdk-py/tree/main#default-authentication-flow to configure credentials for your preferred authentication method."
+            message = f'{message.rstrip(".")}. {default_auth_prompt}'
```
edwardfeng-db, line 621: Added this prompt for cases that debug string is empty while using default auth.

mgyucht (commented):
Great idea, but one suggestion on where to include this.

mgyucht (approved):
Basically LGTM. My only question is whether you think it may be better to refer to our public docs (https://docs.databricks.com/en/dev-tools/auth.html#databricks-client-unified-authentication) or our repo's docs. To be honest, to eliminate having two different doc sources for the SDK, it might be better to link to public docs and later remove some of the documentation from the SDK README.md. What do you think?

tests/test_auth.py
```diff
@@ -4,19 +4,20 @@
 
 from .conftest import __tests__, raises
 
+default_auth_base_error_message = "default auth: cannot configure default credentials, please check https://github.com/databricks/databricks-sdk-py/tree/main#default-authentication-flow to configure credentials for your preferred authentication method"
```
tanmay-db, line 7: Thanks for the PR. I think @hectorcast-db was working on better handling, not sure if there are intersections but mentioning for viz.

Also, if going ahead with this, we would have to do this change in all SDKs for consistency.

hectorcast-db: I was trying to fix the error messages coming from the API to us (which we display to the users). There are probably cases were both messages are displayed, but in general is a different error message.

PR #426, opened 2023-11-06
Make test_auth no longer fail if you have a default profile setup

edwardfeng-db:
## Changes

- Previously `test_auth.py` will have several test cases failing if you have a default profile locally, that's because the sdk code will always look for the `~/.databrickscfg` file, which exists on your machine, but is not expected
- Making specific test cases affected by this behavior use a fake file system instead of the local disk to bypass this issue.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


databricks/sdk/core.py
```diff
@@ -873,7 +873,12 @@ def _known_file_config_loader(self):
             return
         config_file = self.config_file
         if not config_file:
-            config_file = "~/.databrickscfg"
+            if os.environ.get('RUNNING_IN_TEST_MODE') and "testdata" not in os.environ.get('HOME'):
```
edwardfeng-db, line 876: Make it look for a different path, when this is in test mode and `HOME` is not overwritten.

Not sure if this is the best approach. @mgyucht Let me know if you have better ideas

mgyucht (approved):
Nice!

PR #434, opened 2023-11-08
[NOT-MERGED] Support GCP auth

edwardfeng-db:
## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


databricks/sdk/core.py
```diff
@@ -18,8 +18,12 @@
 from typing import (Any, BinaryIO, Callable, Dict, Iterable, Iterator, List,
                     Optional, Type, Union)
 
+import google
+import google.auth
```
nfx, line 22: this would fail in runtime, as `google` package won't be installed. 

can you re-implement google auth the same way we do for Azure? shouldn't be that difficult, there are only 2-3 flows we need to support.

nfx (changes requested):

databricks/sdk/core.py
edwardfeng-db: Sure, just curious why won't they be installed? Don't we have `setup.py` which declares all of the dependencies we wanna install?

edwardfeng-db:
Closing this one, eee https://github.com/databricks/databricks-sdk-py/pull/444 for more status

PR #433, opened 2023-11-08
Adding iterators for SCIM APIs

xinjiezhen-db:
## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied
- [ ] relevant integration tests applied

PR #439, opened 2023-11-14
Update OpenAPI spec for Python SDK

mgyucht:
## Changes
This PR updates the OpenAPI spec for the Python SDK to d136ad0541f036372601bad9a4382db06c3c912d. As part of this, we add some protection against fields or methods in our spec that conflict with reserved keywords in Python.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


databricks/sdk/service/compute.py
```diff
@@ -2106,16 +2115,70 @@ def from_dict(cls, d: Dict[str, any]) -> 'GlobalInitScriptUpdateRequest':
                    script_id=d.get('script_id', None))
 
 
+@dataclass
+class InitScriptEventDetails:
+    cluster: Optional['List[InitScriptInfoAndExecutionDetails]'] = None
+    global_: Optional['List[InitScriptInfoAndExecutionDetails]'] = None
```
mgyucht, line 2121: Here is the corrected rename of the `global` field to `global_`

databricks/sdk/service/iam.py
```diff
@@ -1175,12 +1179,24 @@ def list(self,
...
-        json = self._api.do('GET',
-                            f'/api/2.0/accounts/{self._api.account_id}/scim/v2/Groups',
-                            query=query,
-                            headers=headers)
-        parsed = ListGroupsResponse.from_dict(json).resources
-        return parsed if parsed is not None else []
+
+        # deduplicate items that may have been added during iteration
+        seen = set()
+        query['startIndex'] = 0
```
mgyucht, line 1185: This is incorrect but will be fixed in https://github.com/databricks/databricks-sdk-py/pull/433/files (start index should be 1).

.codegen/service.py.tmpl
```diff
@@ -188,10 +188,10 @@ class {{.Name}}API:{{if .Description}}
         {{template "method-call" .}}
 
     {{if and .Wait (and (not .IsCrudRead) (not (eq .SnakeName "get_run"))) }}
```
tanmay-db, line 190: Thanks for doing this, LGTM, I have one concern -- this might lead to confusion on when we use `.SnakeName` and when `template "safe-snake-name" .`. Should we use safe-snake-name everywhere? 

Longer term, I think we will remove this once we will have spec linter and then it would be one time effort to rename these back to `.SnakeName`

databricks/sdk/service/compute.py
tanmay-db: Nice

tanmay-db (approved):

.codegen/service.py.tmpl
mgyucht: No, you only need to use `template "safe-snake-name" .` when generating a field or method name. There are places where we check whether the operation is `get_run`. In that case, it doesn't matter too much, since that is not a reserved keyword.

With this PR in place, there isn't a need for the linter anymore, since we'll safely rename fields that conflict with keywords.

PR #440, opened 2023-11-14
Paginate all SCIM list requests in the SDK

mgyucht:
## Changes
This PR incorporates two hard-coded changes for the SCIM API in the Python SDK:
1. startIndex starts at 1 for SCIM APIs, not 0. However, the existing .Pagination.Increment controls both the start index as well as whether the pagination is per-page or per-resource. Later, we should replace this extension with two independent OpenAPI options: `one_indexed` (defaulting to `false`) and `pagination_basis` (defaulting to `resource` but can be overridden to `page`).
2. If users don't specify a limit, the SDK will include a hard-coded limit of 100 resources per request. We could add this to the OpenAPI spec as an option `default_limit`, which is useful for any non-paginated APIs that later expose pagination options and allow the SDK to gracefully support those. However, we don't want to encourage folks to use this pattern: all new list APIs are required to be paginated from the start.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


tests/integration/test_iam.py
```diff
@@ -0,0 +1,43 @@
...
+def test_scim_error_unmarshall(w, random):
+    with pytest.raises(DatabricksError) as exc_info:
+        w.groups.list(filter=random(12))
+    assert 'Given filter operator is not supported' in str(exc_info.value)
+
+
+@pytest.mark.parametrize("path,call", [
+    ("/api/2.0/preview/scim/v2/Users", lambda w: w.users.list(count=1)),
+    ("/api/2.0/preview/scim/v2/Groups", lambda w: w.groups.list(count=1)),
+    ("/api/2.0/preview/scim/v2/ServicePrincipals", lambda w: w.service_principals.list(count=1)),
```
nfx, line 21:
```suggestion
    ("/api/2.0/preview/scim/v2/Users", lambda w: w.users.list(count=10)),
    ("/api/2.0/preview/scim/v2/Groups", lambda w: w.groups.list(count=10)),
    ("/api/2.0/preview/scim/v2/ServicePrincipals", lambda w: w.service_principals.list(count=10)),
```

i think count=1 would make integration tests too slow...

tests/integration/test_iam.py
```diff
@@ -0,0 +1,43 @@
...
+def test_workspace_users_list_pagination(w, path, call):
+    raw = w.api_client.do('GET', path)
+    total = raw['totalResults']
+    all = call(w)
+    found = len(list(all))
+    assert found == total
+
+
+@pytest.mark.parametrize("path,call", [
+    # there are ~7k users in our aws prod account
```
nfx, line 32:
```suggestion
```
i don't think it should be in public code

tests/integration/test_iam.py
```diff
@@ -0,0 +1,43 @@
...
+    ("/api/2.0/accounts/%s/scim/v2/Users", lambda a: a.users.list(count=1000)),
+    ("/api/2.0/accounts/%s/scim/v2/Groups", lambda a: a.groups.list(count=1)),
+    # there are ~3k service principals in our aws prod account
```
nfx, line 35:
```suggestion
```

nfx (approved):
thank you for adding this. few nits

databricks/sdk/service/iam.py
```diff
@@ -1182,7 +1182,8 @@ def list(self,
 
         # deduplicate items that may have been added during iteration
         seen = set()
-        query['startIndex'] = 0
+        query['startIndex'] = 1
```
tanmay-db, line 1185: Is this because the first item starts from number 1 in query? https://docs.databricks.com/api/workspace/groups/list

tanmay-db (approved):

tests/integration/test_iam.py
mgyucht: I tried this, but I wanted to make sure that we actually exercised the pagination flows. This test took around 15 seconds for all test cases combined.

tests/integration/test_iam.py
mgyucht: excellent point

tests/integration/test_iam.py
mgyucht: Excellent point

databricks/sdk/service/iam.py
mgyucht: Yes, that's right.

williamdphillips:
Hi @mgyucht & @nfx, maybe I am missing something here but I'm no longer able to query users due to hardcoding of startIndex. 

I have some code that removes inactive users on workspaces, so it iterates over all users. Previously I was using startIndex and count to get 100 users at a time, then query the next 100 users. Now, when I try to query the next 100 users it still shows the first 100 due to this change.

Is this change working as intended?

PR #459, opened 2023-11-28
Adding parameter requirement and sending as body in API call

heimdallur:
## Tests

- [ x ] `make test` run locally
- [ x ] `make fmt` applied


mgyucht:
This file is autogenerated from an API specification, so this change will be overwritten by a future change. Thanks for the contribution, and I'll make sure the autogeneration logic is corrected to address this.

PR #460, opened 2023-11-28
Capture stdout and stderr separately when calling Azure CLI

pietern:
## Changes

They were captured together such that any warnings emitted to stderr would turn the output of the command into malformed JSON and raise an error.

Traces all the way back to #30.

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied


mgyucht (approved):
Thanks!

PR #467, opened 2023-12-01
Reference Documentation Refactoring

mgyucht:
## Changes
This PR makes a number of changes to the documentation with the aim of improving the coverage of reference material as well as making the reference documentation more intuitive to follow by mirroring the structure of the library.

The biggest change here is the architecture: we're no longer generating ReStrucutredText by hand, but rather using the `automodule` directive to generate most of the documentation for us. This means that all classes and enums automatically have documentation, type annotations are present for all parameters, and links and references work properly. The main downside is that examples are no longer included in the main documentation, so I've added a page to the top-level TOC to point out the examples in our repo directly.

To improve the documentation coverage of existing classes, I tweaked some of our templates to include docs from the OpenAPI spec directly in the WorkspaceClient/AccountClient definitions, as well as in the generated code for dataclasses and enums. There are many things that are poorly documented now; we'll need to lean on better linting in the upstream OpenAPI spec to prevent this from getting worse.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


.codegen.json
```diff
@@ -26,7 +26,6 @@
       "pip install '.[dev]'"
     ],
     "post_generate": [
-      "pytest -m 'not integration' --cov=databricks --cov-report html tests",
```
mgyucht, line 29: add this back in

docs/gen-client-docs.py
```diff
@@ -131,129 +69,58 @@ class Generator:
...
 
 .. toctree::
-  :maxdepth: 1
+   :maxdepth: 1
+   
+{all}
+''')
 
-  {all}''')
+    def write_client_reference(self, title: str, class_name: str, filename: str):
```
mgyucht, line 112: Remove

docs/gen-client-docs.py
```diff
@@ -28,31 +20,9 @@ class Tag:
...
-            args = f'{args} [, {other}]'
-        return args
-
-    def as_rst(self, usage) -> str:
-        if self.doc is None: return ''
-        out = ['', f'    .. py:method:: {self.method_name}({self.argspec()})', usage, f'        {self.doc}']
-        return "\n".join(out)
-
-
 @dataclass
```
mgyucht, line 23: remove

pietern (approved):
Really nice to have inline docs for the methods and data classes.

strongbad03:
Ah, this explains why pages like the [Group page](https://databricks-sdk-py.readthedocs.io/en/latest/workspace/groups.html) that I was using last week have since evaporated. I see [its replacement](https://databricks-sdk-py.readthedocs.io/en/latest/autogen/iam.html#databricks.sdk.service.iam.GroupsAPI), but I echo the concern with the remark above that losing examples is definitely a downside.

PR #473, opened 2023-12-05
Use "fromisoformat" to parse tokens from the CLI

fjakobs:
## Changes
Make date parsing for tokens from the CLI more robust

## Tests
* Added unit test
* Manually tested with CLI auth provider


tests/test_core.py
```diff
@@ -52,6 +52,12 @@ def test_databricks_cli_token_source_not_installed(config, monkeypatch):
         DatabricksCliTokenSource(config)
 
 
+def test_databricks_cli_token_parse_expiry():
+    CliTokenSource._parse_expiry("2023-12-01T15:19:48.007742617Z")
+    CliTokenSource._parse_expiry("2023-12-05T15:59:01.40081+11:00")
+    CliTokenSource._parse_expiry("2023-12-06 10:06:05")
```
mgyucht, line 58: Can you just add an assertion on the actual returned values?
BTW: good practice here would be to use `pytest.mark.parameterize` annotation, e.g.
```python
@pytest.mark.parametrize(
    "date_string,expected",
    [("2023-12-01T15:19:48.007742617Z", <the expected date>),
      ("2023-12-05T15:59:01.40081+11:00", <the expected date>),
      ("2023-12-06 10:06:05", <the expected date>)])
def test_databricks_cli_token_parse_expiry(date_string, expected):
```

mgyucht (approved):
Small nit on the tests but otherwise LGTM.

PR #476, opened 2023-12-11
Add MANAGED_SHALLOW_CLONE to TableType Enum

sandormatyas:
## Changes

Adds a new value to `TableType` - `MANAGED_SHALLOW_CLONE` is a table type in Databricks that's currently not present in the `TableType` Enum.

I was not able to find documentation on this (aside https://docs.databricks.com/en/delta/clone-unity-catalog.html), but here's an API call that fetches a table in a schema with the type `MANAGED_SHALLOW_CLONE`:
image

## Tests

I made an attempt to run tests, but some of them fail on my main branch before any changes, so I couldn't reliably run them.

- [ ] `make test` run locally
- [ ] `make fmt` applied


sandormatyas:
Replaced by #632

Issue #343, opened 2023-09-12
SyntaxError: EOL while scanning string literal


ConstantinoSchillebeeckx:
**Description**

We intermittently see the following error:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>

  File "/usr/lib/python3.9/pkgutil.py", line 108, in walk_packages

    yield from walk_packages(path, info.name+'.', onerror)
  File "/usr/lib/python3.9/pkgutil.py", line 108, in walk_packages

    yield from walk_packages(path, info.name+'.', onerror)

  File "/usr/lib/python3.9/pkgutil.py", line 93, in walk_packages

    __import__(info.name)
  File "/databricks/python3/lib/python3.9/site-packages/dbt/adapters/databricks/__init__.py", line 1, in <module>

    from dbt.adapters.databricks.connections import DatabricksConnectionManager  # noqa

  File "/databricks/python3/lib/python3.9/site-packages/dbt/adapters/databricks/connections.py", line 47, in <module>

    from databricks import sql as dbsql
  File "/databricks/python3/lib/python3.9/site-packages/databricks/__init__.py", line 3

    __path__ = __import__('pkgutil').extend_path(__path__, __name__)").extend_path(__path__, __name__)
                                                                                                      ^
SyntaxError: EOL while scanning string literal
```

We've seen this mostly during our CI where we build our Docker container, and as a sanity check, we import our code to ensure everything is kosher. Most recently, we saw this happen as we were running `pytest`.

Clearly the identified line seems to get modified as it differs from the referenced one of:
https://github.com/databricks/databricks-sdk-py/blob/cef100c3a3c9dce91bc2bc8d6c59a93febd8f707/databricks/__init__.py#L3

**Reproduction**

Unfortunately, I'm struggling to reproduce this 😢 

**Expected behavior**
No SyntaxError

**Debug Logs**
Can't generate since I can't reproduce.

**Other Information**
 - OS: Ubuntu 20.04.6
 - Version: 0.6.0

pietern:
Thanks for reporting. It seems something in your setup is adding a double quote to the line producing an error.

Our version of `__init__.py` looks like this:
https://github.com/databricks/databricks-sdk-py/blob/986d1d98d4fea66c99d0ea6ccfc64b9faa1115db/databricks/__init__.py#L3

The error you mention includes a double quote in that line, thus produces the error "EOL while scanning string literal".

Closing because this is not an SDK issue.

[...]

PR #488, opened 2023-12-21
Make contents of `__init__.py` equal across projects

pietern:
## Changes

See https://github.com/databricks/databricks-sdk-py/issues/343#issuecomment-1866029118.

Corresponding PR in the SQL connector: https://github.com/databricks/databricks-sql-python/pull/304

## Tests

Unit test to confirm the SHA1 of the file contents.


mgyucht (approved):

pietern:
Let's hold off on merging until we can make sure we can make the contents for Databricks Connect equal as well.

Issue #494, opened 2024-01-03
`workload_size` should expect a `str`, not an object which references `value` 


jbrill:
https://github.com/databricks/databricks-sdk-py/blame/193c3fbc255262de506b9bb927c87392c2c26720/databricks/sdk/service/serving.py#L1654

When attempting to call with an example:
```
ServedModelInput(
            model_name=model_name,
            model_version=latest_model_version,
            workload_size="Small",
            scale_to_zero_enabled=True,
            environment_vars={}
        )
```

We receive:
```
   1652 if self.name is not None: body['name'] = self.name
   1653 if self.scale_to_zero_enabled is not None: body['scale_to_zero_enabled'] = self.scale_to_zero_enabled
-> 1654 if self.workload_size is not None: body['workload_size'] = self.workload_size.value
   1655 if self.workload_type is not None: body['workload_type'] = self.workload_type.value
   1656 return body

AttributeError: 'str' object has no attribute 'value'
```

mgyucht:
@jbrill can you use `ServedModelInputWorkloadSize.SMALL` for now?

We plan to make the SDK accept strings for enum values in the coming future.

PR #495, opened 2024-01-03
[hotfix] workload size string assignment

jbrill:
Resolves https://github.com/databricks/databricks-sdk-py/issues/494


mgyucht (changes requested):
This is an autogenerated file, and your change will be removed during the next release of the SDK.

We are planning some work to make the Python SDK accept strings in place of enums and objects in place of dataclasses, but that will be a global change that applies to all APIs in the SDK.

jbrill:
> This is an autogenerated file, and your change will be removed during the next release of the SDK.
> 
> We are planning some work to make the Python SDK accept strings in place of enums and objects in place of dataclasses, but that will be a global change that applies to all APIs in the SDK.

Sounds good. I was able to circumvent the issue temporarily by passing an object directly to the `workload_size` parameter:
```
class WorkloadSize:
    def __init__(self, value):
        self.value = value

workload_size = WorkloadSize("Small")
```

renaudhartert-db:

PR #506, opened 2024-01-11
Don't throw `TimeoutError` but rather the last error of retries

nfx:
## Changes
If we `raise TimeoutError(..) from last_err`, pytest displays the timeout error in the summary `FAILED ...::test_running_real_assessment_job - TimeoutError: Timed out after 0:06:00`, which makes troubleshooting harder during the stress time.

This change will put TimeoutError in the __cause__ attribute of re-raised last error. See more at https://docs.python.org/3/library/exceptions.html#exception-context

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/databricks/sdk/retries.py
+++ b/databricks/sdk/retries.py
@@ -52,7 +52,14 @@ def retried(*,
                     logger.debug(f'Retrying: {retry_reason} (sleeping ~{sleep}s)')
                     time.sleep(sleep + random())
                     attempt += 1
-            raise TimeoutError(f'Timed out after {timeout}') from last_err
+
+            # If we `raise TimeoutError(..) from last_err`, pytest displays the timeout error in the summary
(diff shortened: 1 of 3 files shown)


jasongrout-db:
This seems like it reverses the conventional order of chained exceptions. Is this change just to accommodate pytest's printing? In general, I would suggest not changing the "correct" user-facing behavior to accommodate testing infrastructure.

It seems that pytest has had a lot of work recently in improving behavior and display of nested exceptions and exception groups. Perhaps the simplest is manually checking the cause, like in 
[this pytest issue, showing how to catch and test these types of exceptions](https://github.com/pytest-dev/pytest/issues/6694). Would that help? Are you testing with the most recent pytest? I see there were very recent changes from late last year around exception handling.

hectorcast-db:
Errors have been rewritten. Closing this PR.

PR #514, opened 2024-01-19
Support dev and staging workspaces

mgyucht:
## Changes
Copies the DatabricksEnvironment abstraction from the Go SDK to the Python SDK. This enables using the SDK for Azure development and staging workspaces, which use an alternate Azure login application ID.

Additionally, this PR changes `azure_environment` to accept the same values as the Go SDK and Terraform providers for both Databricks and Azure, i.e. `PUBLIC`, `USGOVERNMENT`, and `CHINA`.

The first commit of this PR is a refactor moving config & credentials provider code to separate files, as core.py is getting quite big and difficult to maintain. The second one is much smaller and has just the net new bits.

## Tests
- [x] Unit test for the new environment lookup code.
- [x] Manual test Azure CLI auth against a staging Azure workspace.


tests/test_core.py
```diff
@@ -282,9 +285,10 @@ class DummyResponse(requests.Response):
     _closed: bool = False
 
     def __init__(self, content: List[bytes]) -> None:
+        super().__init__()
         self._content = iter(content)
 
-    def iter_content(self, chunk_size: int = 1) -> Iterator[bytes]:
+    def iter_content(self, chunk_size: int = 1, decode_unicode=False) -> Iterator[bytes]:
```
mgyucht, line 291: This fixes some yellow squigglies in IntelliJ.

databricks/sdk/core.py
```diff
@@ -1,1040 +1,24 @@
...
-from .azure import (ARM_DATABRICKS_RESOURCE_ID, ENVIRONMENTS, AzureEnvironment,
-                    add_sp_management_token, add_workspace_id_header)
+from .config import Config
 from .errors import DatabricksError, error_mapper
-from .oauth import (ClientCredentials, OAuthClient, OidcEndpoints, Refreshable,
-                    Token, TokenCache, TokenSource)
 from .retries import retried
-from .version import __version__
 
 __all__ = ['Config', 'DatabricksError']
```
nfx, line 18: i'm relying on `credentials_provider` and couple of other classes from `core`. don't break it. Please verify that https://github.com/databrickslabs/ucx still works before this PR is merged. If you make a PR to accommodate these changes - also fine

nfx (changes requested):

databricks/sdk/core.py
mgyucht: Just added imports in `core.py` to make sure all credentials_provider.py and config.py classes are reexported. Manually tested and was able to import those definitions after installing.

databricks/sdk/environments.py
```diff
@@ -0,0 +1,72 @@
...
+            return None
+        return self.azure_environment.active_directory_endpoint
+
+
+DEFAULT_ENVIRONMENT = DatabricksEnvironment(Cloud.AWS, ".cloud.databricks.com")
+
+ALL_ENVS = [
+    DatabricksEnvironment(Cloud.AWS, ".dev.databricks.com"),
+    DatabricksEnvironment(Cloud.AWS, ".staging.cloud.databricks.com"),
+    DatabricksEnvironment(Cloud.AWS, ".cloud.databricks.us"), DEFAULT_ENVIRONMENT,
```
hectorcast-db, line 48: Nit, put DEFAULT_ENVIRONMENT in a new line (maybe in the first line).

databricks/sdk/azure.py
```diff
@@ -15,19 +15,15 @@ class AzureEnvironment:
 ARM_DATABRICKS_RESOURCE_ID = "2ff814a6-3304-4ab8-85cb-cd0e6f879c1d"
 
 ENVIRONMENTS = dict(
-    PUBLIC=AzureEnvironment(name="AzurePublicCloud",
+    PUBLIC=AzureEnvironment(name="PUBLIC",
```
hectorcast-db, line 18: From what I see, the name is not really used. Correct?

databricks/sdk/azure.py
```diff
@@ -15,19 +15,15 @@ class AzureEnvironment:
...
                             service_management_endpoint="https://management.core.windows.net/",
                             resource_manager_endpoint="https://management.azure.com/",
                             active_directory_endpoint="https://login.microsoftonline.com/"),
-    GERMAN=AzureEnvironment(name="AzureGermanCloud",
-                            service_management_endpoint="https://management.core.cloudapi.de/",
-                            resource_manager_endpoint="https://management.microsoftazure.de/",
-                            active_directory_endpoint="https://login.microsoftonline.de/"),
-    USGOVERNMENT=AzureEnvironment(name="AzureUSGovernmentCloud",
+    USGOVERNMENT=AzureEnvironment(name="USGOVERNMENT",
```
hectorcast-db, line 22: Why do we remove the German cloud?

databricks/sdk/environments.py
mgyucht: This was done by our formatter, so I think I wil leave this as is.

databricks/sdk/azure.py
mgyucht: This was added unnecessarily. Databricks is not run in the German cloud, and we don't include it in the [Go](https://github.com/databricks/databricks-sdk-go/blob/main/config/azure.go#L122) SDK either. It is purely in maintenance mode since 2018: https://learn.microsoft.com/en-us/azure/germany/.

hectorcast-db (approved):

databricks/sdk/azure.py
mgyucht: In general, users don't need to specify this if they specify the host name, since we can derive the Azure environment from the hostname. However, if users only specify the workspace resource ID and are deploying in either China or Govcloud, they need to specify either `ARM_ENVIRONMENT` or the `azure_environment` parameter in WorkspaceClient/AccountClient/Config.

This PR does change this (before you would specify `AzurePublicCloud` and now you would specify `PUBLIC` or `public`), aligning with the [Azure TF provider](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#environment) and with the [Go SDK](https://github.com/databricks/databricks-sdk-go/blob/main/config/azure.go#L123).

mgyucht: Follow-up: I added back compatibility for the old style of environment names just in case, so customers specifying these parameters now will not need to change.

PR #517, opened 2024-01-25
Search for both databricks.exe and databricks binaries in windows

kartikgupta-db:
## Tests

- [x] `make test` run locally
- [x] `make fmt` applied


mgyucht (approved):
LGTM, one small nit

PR #530, opened 2024-02-06
Do not terminate listing for token-based pagination resources on empty response

mgyucht:
## Changes
It is allowed for list APIs to return less than the number of resources in the page_size limit in a response. In case no items are returned but there are still items to list, we should continue to list resources until no next_page_token is included in the response. This matches the logic from the Go SDK: https://github.com/databricks/databricks-sdk-go/blob/main/.codegen/api.go.tmpl#L275-L294

## Tests
- [x] Nightly tests pass.


nfx (approved):
lgtm

PR #532, opened 2024-02-07
Fix get_workspace_client in GCP

mgyucht:
## Changes
The current implementation of get_workspace_client copies the entire config, critically reusing the cached header factory so as to use the same auth mechanism when getting a workspace client. However, at least in GCP, account-level OAuth tokens can't be used to authenticate to a workspace (probably because the audience for the account-level and workspace-level tokens is different).

This PR fixes this by only copying the exported fields and not copying the header factory. Subsequent use of the config in WorkspaceClient will trigger config resolution. For GCP, this means creating a new token source using the correct host as the audience.

This ports https://github.com/databricks/databricks-sdk-go/pull/803 to the Python SDK.

## Tests
Manually ran this integration test in all non-UC (Azure, AWS, GCP) and UC (AWS, GCP) account-level environments.

- [ ] `make test` run locally
- [ ] `make fmt` applied


.codegen/__init__.py.tmpl
```diff
@@ -146,10 +146,17 @@ class AccountClient:
         :param workspace: The workspace to construct a client for.
         :return: A ``WorkspaceClient`` for the given workspace.
         """
-        config = self._config.copy()
-        config.host = config.environment.deployment_url(workspace.deployment_name)
-        config.azure_workspace_resource_id = azure.get_azure_resource_id(workspace)
-        config.account_id = None
+        config_inner = self._config.as_dict().copy()
+        if 'host' in config_inner:
+            del config_inner['host']
```
tanmay-db, line 151: using something like:
`keys_to_remove = ['host', 'azure_workspace_resource_id', 'account_id']`

might be better for readability / extensible in case we want to have more keys we want to remove in future.

tanmay-db (approved):

PR #533, opened 2024-02-08
Use fake clock for faster unit tests

mgyucht:
## Changes
Some unit tests of retry logic are slow because they sleep for a second or two. This PR changes ApiClient to accept a Clock which supplies the time for retries and waiting. In unit tests, we use a fake clock that doesn't tick until sleep() is called, and then ticks by the amount provided.

## Tests
Existing unit tests.

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/databricks/sdk/clock.py
+++ b/databricks/sdk/clock.py
@@ -1,0 +1,50 @@
+import abc
+import time
+
+
+class Clock(metaclass=abc.ABCMeta):
+    @abc.abstractmethod
+    def time(self) -> float:
+        """
+        Return the current time in seconds since the Epoch.
+        Fractions of a second may be present if the system clock provides them.
+
+        :return: The current time in seconds since the Epoch.
+        """
+        pass
+
+    @abc.abstractmethod
+    def sleep(self, seconds: float) -> None:
+        """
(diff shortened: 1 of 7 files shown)


databricks/sdk/clock.py
```diff
@@ -0,0 +1,49 @@
...
+
+        :return: The current time in seconds since the Epoch.
+        """
+        return time.time()
+
+    def sleep(self, seconds: float) -> None:
+        """
+        Return the current time in seconds since the Epoch.
+        Fractions of a second may be present if the system clock provides them.
+
```
edwardfeng-db, line 45: These lines don't seem to be accurate?

databricks/sdk/clock.py
```diff
@@ -0,0 +1,49 @@
...
+        :return: The current time in seconds since the Epoch.
+        """
+
+    @abc.abstractmethod
+    def sleep(self, seconds: float) -> None:
+        """
+        Return the current time in seconds since the Epoch.
+        Fractions of a second may be present if the system clock provides them.
+
```
edwardfeng-db, line 21: Don't seem to be accurate?

edwardfeng-db (approved):

PR #547, opened 2024-02-16
Support HEAD operation and response Headers

hectorcast-db:
## Changes
Support HEAD operation and response Headers 

## Test

- [X] make fmt
- [X] make test
- [X] run integration tests (both with current sha and master sha)


.codegen/__init__.py.tmpl
```diff
@@ -153,7 +153,7 @@ class AccountClient:
         :param workspace: The workspace to construct a client for.
         :return: A ``WorkspaceClient`` for the given workspace.
         """
-        config = self._config.copy()
+        config = self._config.deep_copy()
```
hectorcast-db, line 156: Bug fix. The generated code was already updated here

https://github.com/databricks/databricks-sdk-py/commit/9af2630672c866610248243eb53898c16c452960#diff-3847ef3d682f5902da2252a8d7b3a0745d15a4b81ef42cd73f73be158622c5d6R812

.codegen/service.py.tmpl
```diff
@@ -21,26 +21,26 @@ from databricks.sdk.service import {{.Package.Name}}{{end}}
 {{if .Fields -}}{{if not .IsRequest}}@dataclass
 class {{.PascalName}}{{if eq "List" .PascalName}}Request{{end}}:{{if .Description}}
     """{{.Comment "    " 100}}"""
-    {{end}}{{- range .RequiredFields | alphanumOnly}}
+    {{end}}{{- range .RequiredFields}}
```
hectorcast-db, line 24: header fields are not alphanumeric

tests/integration/test_files.py
```diff
@@ -221,7 +221,9 @@ def test_files_api_upload_download(ucws, random):
             f = io.BytesIO(b"some text data")
             target_file = f'/Volumes/main/{schema}/{volume}/filesit-{random()}.txt'
             w.files.upload(target_file, f)
-
+            # TODO: Enable after generating with the latest spec
```
hectorcast-db, line 224: Added to run integration tests with the master sha. I will uncomment it later today during the release.

mgyucht (commented):
The ApiClient.do() interface is really hard to work around. It assumes too much about how requests are deserialized, resulting in really strange workarounds like this. We do need to revisit this, but it may be a bit more work than we can afford given our current timelines.

One idea would be to introduce a "Response" wrapper type like we have in the Go/Java SDKs. This would expose the core response attributes, like `json()` to deserialize the response as JSON, headers, etc. In the other SDKs, the type is a parameter to the ApiClient, meaning that we can deserialize reflectively in the API client depending on what is requested. Again, this isn't fleshed out, so we would need to work through it.

In the meantime, I think we should try to preserve some consistency by making `from_dict` work well with the response of `ApiClient.do` without such a drastic change to the interface.

databricks/sdk/core.py
```diff
@@ -135,11 +135,14 @@ def do(self,
...
-        if raw:
-            return StreamingResponse(response)
+
         resp = dict()
         for header in response_headers if response_headers else []:
             resp[header] = response.headers.get(Casing.to_header_case(header))
+        if raw:
+            # Don't add the StreamingReponse to the dict.
+            # from_dict does not support it.
+            return resp, StreamingResponse(response)
```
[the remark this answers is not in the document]
hectorcast-db: Fixed

.codegen/service.py.tmpl
```diff
@@ -304,12 +313,10 @@ class {{.Name}}API:{{if .Description}}
 {{- end}}
 
 {{define "method-call-default" -}}
-        {{if .Response -}}
+        {{if and .Response -}}
```
mgyucht, line 316:
```suggestion
        {{if .Response -}}
```

mgyucht (approved):
One small nit, otherwise LGTM. This is worth calling out in the release notes about the change in behavior of the `raw` flag on `ApiClient.do()`.

PR #559, opened 2024-02-22
Support subservices

hectorcast-db:
## Changes
Support subservices
Test PR with custom OpenAPI spec
https://github.com/databricks/databricks-sdk-py/pull/558

Docs for linked PR
https://databricks-sdk-py--558.org.readthedocs.build/en/558/workspace/settings/index.html#

## Tests

- [X] `make test` run locally
- [X] `make fmt` applied
- [X] relevant integration tests applied


mgyucht (commented):
This looks good. Let's make sure we get the changes needed in gen-client-docs.py into this PR as well.

docs/gen-client-docs.py
```diff
@@ -54,6 +54,23 @@ def __str__(self):
...
+
+    def as_rst(self) -> str:
+        out = ['', f'    .. py:property:: {self.name}']
+        if self.tpe is not None:
+            out.append(f'        :type: {self.tpe}')
+        if self.doc is not None:
+            # This is a class doc, which comes with 4 indentation spaces. 
+            # Here we are using the doc as property doc, which needs 8 indentation spaces.
+            formatted_doc = re.sub(r'\n', '\n    ', self.doc)
+            out.append(f'\n        {formatted_doc}')
```
mgyucht, line 71: I added `_get_enum_doc` recently that basically does this. Let's see if we can refactor/reuse that method?

docs/gen-client-docs.py
```diff
@@ -275,6 +296,23 @@ def _to_typed_args(argspec: inspect.FullArgSpec, required: bool) -> list[TypedAr
...
+    def class_properties(self, inst) -> list[PropertyDoc]:
+        property_docs = []
+        for name in dir(inst):
+            if name[0] == '_':
+                # private members
+                continue
+            instance_attr = getattr(inst, name)
+            if name.startswith('_'):
+                continue
+            if inspect.ismethod(instance_attr):
```
mgyucht, line 308: Can we try this? It is probably a bit more robust.
```suggestion
            if not isinstance(inst, property):
```

docs/gen-client-docs.py
```diff
@@ -275,6 +296,23 @@ def _to_typed_args(argspec: inspect.FullArgSpec, required: bool) -> list[TypedAr
...
+                continue
+            property_docs.append(
+                PropertyDoc(name=name,
+                            doc=instance_attr.__doc__,
+                            tpe=instance_attr.__class__.__name__ if instance_attr.__class__ != property else None))
```
mgyucht, line 313: When is the `else` branch taken?

docs/conf.py
```diff
@@ -5,6 +5,8 @@
 
 # -- Project information -----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+import sys
```
mgyucht, line 8: Does `make -C docs livehtml` work when run from the root of the repo?

mgyucht (approved):
Overall, looks good to me. Just a couple small pieces of feedback.

docs/conf.py
hectorcast-db: Yes

docs/gen-client-docs.py
hectorcast-db: From a reflection point of view, the exact object we have here is not a property.

docs/gen-client-docs.py
hectorcast-db: Not needed. It was needed with the first version which was not filtering everything correctly.

docs/gen-client-docs.py
hectorcast-db: That method also strips double returns to keep everything as a single paragraph.

PR #560, opened 2024-02-22
Sort index pages by name in docs

mgyucht:
## Changes
The indices under the APIs and Dataclasses pages were not sorted, resulting in them being shown in an unexpected/random order. This fixes this so they are shown alphabetically.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/docs/account/index.rst
+++ b/docs/account/index.rst
@@ -7,9 +7,9 @@ These APIs are available from AccountClient
 .. toctree::
    :maxdepth: 1
 
-   iam/index
-   catalog/index
-   settings/index
-   provisioning/index
    billing/index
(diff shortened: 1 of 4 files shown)


tanmay-db (approved):
I see we are doing:
````
doc = DataclassesDoc(package=pkg, dataclasses=sorted(all_members))
            with open(f'{__dir__}/dbdataclasses/{pkg.name}.rst', 'w') as f:
                f.write(doc.as_rst())
````

Do you think we should sort this as well?
````
        for d in self.dataclasses:
            out.append(self.dataclass_rst(d))
        return "\n".join(out)
````

PR #561, opened 2024-02-22
Distinguish between empty types and fields that can take any value

mgyucht:
## Changes
Ports https://github.com/databricks/databricks-sdk-go/pull/821 to the Python SDK.

This also bumps the API spec to the same version as in the above PR.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


.codegen/service.py.tmpl
```diff
@@ -71,46 +79,31 @@ class {{.PascalName}}{{if eq "List" .PascalName}}Request{{end}}:{{if .Descriptio
...
-	{{- else if .IsInt}}int
-	{{- else if .IsByteStream}}BinaryIO
+	{{- if .IsEmpty}}:class:`{{template "type" .}}`
 	{{- else if .ArrayValue }}List[{{template "type-doc" .ArrayValue}}]
 	{{- else if .MapValue }}Dict[str,{{template "type-doc" .MapValue}}]
 	{{- else if .IsExternal }}:class:`{{.PascalName}}`
 	{{- else if .IsObject }}:class:`{{.PascalName}}{{if eq "List" .PascalName}}Request{{end}}`
 	{{- else if .Enum }}:class:`{{.PascalName}}`
-	{{- else}}any /* MISSING TYPE */
+	{{- else}}{{template "type" . }}
```
mgyucht, line 106: This removes unnecessary duplication between type and type-doc, and removes the completely duplicated type-nq template. There is no behavior change here, other than changing `.IsEmpty` from `Any` to `{{.PascalName}}`.

hectorcast-db (approved):

PR #562, opened 2024-02-22
Make a best effort attempt to initialise all Databricks globals

kartikgupta-db:
## Changes
We only initialise dbutils locally when using `from databricks.sdk.runtime import *`. Users in the webui are guided to use this import in all library code. 
image

The local (for people outside DBR) solution so far was to initialise spark manually. But this can be tedious for deeply nested libraries (which is the reason this import was introduced in the first place). 

Now, we make a best effort attempt to initialise maximum number of globals locally, so that users can build and debug libraries using databricks connect. 

## Tests
* integration test

- [ ] `make test` run locally
- [x] `make fmt` applied


databricks/sdk/runtime/__init__.py
```diff
@@ -86,23 +86,87 @@ def inner() -> Dict[str, str]:
...
     is_local_implementation = True
 
+    try:
+        # We expect this to fail and only do this for providing types
+        from pyspark.sql.context import SQLContext
+        sqlContext: SQLContext = None # type: ignore
+        sql = sqlContext.sql
+        table = sqlContext.table
+    except Exception:
+        pass
```
mgyucht, line 99: suggestion: debug log here.

databricks/sdk/runtime/__init__.py
```diff
@@ -86,23 +86,87 @@ def inner() -> Dict[str, str]:
...
+        logging.debug(f"Failed to initialise udf global: {e}")
+
+    try:
+        from databricks.connect import DatabricksSession  # type: ignore
+        spark = DatabricksSession.builder.getOrCreate()
+        sc = spark.sparkContext
+    except Exception as e:
+        # We are ignoring all failures here because user might want to initialize
+        # spark session themselves and we don't want to interfere with that
+        logging.debug(f"Failed to initialize spark session: {e}")
```
mgyucht, line 115: in these debug messages, let's be explicit about which globals were not defined, but also indicate that it wasn't a problem.
```suggestion
        logging.debug(f"Failed to initialize globals 'spark' and 'sc', continuing. Cause: {e}")
```

databricks/sdk/runtime/__init__.py
```diff
@@ -86,23 +86,87 @@ def inner() -> Dict[str, str]:
...
+                            If data is a URL, display the resource at that URL, the resource is loaded dynamically by the browser.
+                            Otherwise data should be the HTML to be displayed.
+            See also:
+            IPython.display.HTML
+            IPython.display.display_html
+            """
+            return IPDisplay.display_html(html, raw=True) # type: ignore
+
+    except ImportError as e:
+        logging.debug(f"Failed to initialise display globals: {e}")
```
mgyucht, line 157:
```suggestion
        logging.debug(f"Failed to initialise globals 'display' and 'displayHTML', continuing. Cause: {e}")
```

tests/integration/test_local_globals.py
```diff
@@ -0,0 +1,30 @@
...
+@contextmanager
+def restorable_env():
+    import os
+    current_env = os.environ.copy()
+
+    try:
+        yield
+    finally:
+        os.environ.clear()
+        os.environ.update(current_env)
```
mgyucht, line 13: All tests will share the same environment, right? This could cause issues for tests running concurrently that also inspect the process environment, especially clearing the environment. To minimize the risk of other tests missing some information from the environment, what do you think of:
```suggestion
        for k, v in os.environ.enumerate():
            if k not in current_env:
                del os.environ[k]
            else if v != current_env[k]:
                os.environ[k] = current_env[k]
```

mgyucht (approved):
Thanks for this! Few suggestions, but looks good overall. Can merge once these comments are addressed.

fjakobs (approved):

mgyucht (approved):
Thanks, LGTM with two small nits.

PR #568, opened 2024-02-28
Add `with_library(product, version)` to track usage of each upstream library separately

nfx:
This PR adds `with_library(product, version)` method to clients & config, creating a copy with the user-agent extra propagated only for the relevant calls


nfx:
superceded by https://github.com/databricks/databricks-sdk-py/pull/679

PR #570, opened 2024-03-01
Fix QueryHistoryAPI.list()

MBosmanABN:
## Changes
Adds additional check for same `page_token` because that seems to have been an issue causing it to hang in our environment.

## Tests
I didn't, except for running it on our environment and I'm too lazy to add proper tests.

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/databricks/sdk/service/sql.py
+++ b/databricks/sdk/service/sql.py
@@ -4702,7 +4702,7 @@ class QueryHistoryAPI:
             if 'res' in json:
                 for v in json['res']:
                     yield QueryInfo.from_dict(v)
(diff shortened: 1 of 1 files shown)


mgyucht:
Hi @MBosmanABN, thanks for making this PR.

The backend team for query history has just fixed the underlying issue (they will stop including the next page token on the last page response), so it will be fixed in about a week as it is rolled out to all regions, and it won't require any changes in SDKs. Separately, these files are autogenerated, so changes will be overwritten. Consequently, I will close this PR.

PR #572, opened 2024-03-02
New example to list compute resource for SUBMIT_RUN job runs

kimberlyma:
## Changes
From a discussion with @mgyucht on a request for getting a list of jobs submitted via jobs/submit and the associated compute resource(s). There are multiple compute-related fields returned in the response (both on the run, and on the task) so this example aims to disambiguate them.

Ran against AWS, Azure, GCP workspaces.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/examples/list_compute_submitrun_runs.py
+++ b/examples/list_compute_submitrun_runs.py
@@ -1,0 +1,42 @@
+#!env python3
+import logging
+import sys
+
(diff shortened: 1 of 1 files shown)


mgyucht (approved):
Awesome, thanks!

PR #576, opened 2024-03-06
Update SDK to latest OpenAPI spec

tanmay-db:
## Changes

- Update SDK to OpenAPI spec
- Also, fix naming convention, we use Pascal case (.PascalName) instead of Camel case (.Name). Because of this we now generate the following:
`CSPEnablement` -> `CspEnablement`

## Tests

Unit tests, integration test will run release PR. 
- [ ] `make test` run locally
- [ ] `make fmt` applied


.codegen/service.py.tmpl
```diff
@@ -101,7 +101,7 @@ class {{.PascalName}}{{if eq "List" .PascalName}}Request{{end}}:{{if .Descriptio
 {{- end -}}
 
 {{range .Services}}
-class {{.Name}}API:{{if .Description}}
+class {{.PascalName}}API:{{if .Description}}
```
tanmay-db, line 104: Highlight

docs/gen-client-docs.py
```diff
@@ -388,7 +388,7 @@ def _get_tag_name(self, class_name, service_name) -> Tag:
             class_name = class_name.replace('Ext', 'API')
         class_name = class_name[:-3]
         for tag_name, t in self.mapping.items():
-            if t.service == class_name:
+            if t.service.lower() == str(class_name).lower():
```
tanmay-db, line 391: Highlight

mgyucht: Note that the SDK has tools to manipulate string casing: https://github.com/databricks/databricks-sdk-py/blob/main/databricks/sdk/casing.py. However, I think this is still safe enough as we shouldn't have services that are the same modulo case.

.codegen/service.py.tmpl
mgyucht: OK, as this is aligned with the treatment of this in Go SDK.

mgyucht (approved):
LGTM. We need to announce in this release the change to the settings APIs.

PR #579, opened 2024-03-07
Fix null body response to empty in ApiClient

tanmay-db:
## Changes

Some delete end points return null instead of empty body, we fix that in SDK Client since the APIs are not available publicly and support isn't available for them yet. 

## Tests

Thanks to @sodle-splunk for contributing the unit test (https://github.com/databricks/databricks-sdk-py/pull/641)
- [x] `make test` run locally
- [x] `make fmt` applied


mgyucht (commented):
Test?

tanmay-db:
> Test?

@mgyucht, I have asked on slack for confirmation on the fix, since the API is private, we might need to enable it for our test infra and then repro this, it would be some work, once confirmed I will add unit test. Me and @edwardfeng-db are working on other issues in the meantime parallelly.

tanmay-db:
Note: The failing test: `downstreams / compatibility (blueprint, databrickslabs) (pull_request)` fails on a simple text change: https://github.com/databricks/databricks-sdk-py/pull/643. This passed on last commit which was 2 weeks ago and doesn't seem to be happening due to changes in Python SDK.

nkvuong:
the failed check is not related to this change, the same test is failing on blueprint/main

sodle-splunk (approved):
Thanks all for quick response and feedback on this one. This bug is one of our final blockers to shipping a new integration. 🥳

tanmay-db:
Hi @sodle-splunk, thanks for raising this, the release with fix has been done: https://github.com/databricks/databricks-sdk-py/releases/tag/v0.27.1

PR #591, opened 2024-03-22
Override INVALID_PARAMETER_VALUE on fetching non-existent job/cluster

mgyucht:
## Changes
Port of https://github.com/databricks/databricks-sdk-go/pull/864 to the Python SDK.

Most services use RESOURCE_DOES_NOT_EXIST error code with 404 status code to indicate that a resource doesn't exist. However, for legacy reasons, Jobs and Clusters services use INVALID_PARAMETER_VALUE error code with 400 status code instead. This makes tools like Terraform and UCX difficult to maintain, as these services need different error handling logic. However, we can't change these behaviors as customers already depend on the raw HTTP response status & contents.

This PR corrects these errors in the SDK itself. SDK users can then do

```
try:
    w.jobs.get_by_id('abc')
except ResourceDoesNotExist:
    pass
```
just as you would expect from other resources.

Updated the README with more information about this as well.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/databricks/sdk/errors/mapper.py
+++ b/databricks/sdk/errors/mapper.py
@@ -1,8 +1,72 @@
 from databricks.sdk.errors import platform
 from databricks.sdk.errors.base import DatabricksError
+from dataclasses import dataclass
+import re
+from typing import Optional
+import requests
+from .platform import ResourceDoesNotExist
 
 
-def error_mapper(status_code: int, raw: dict) -> DatabricksError:
+@dataclass
+class _ErrorOverride:
+    # The name of the override. Used for logging purposes.
+    debug_name: str
+
+    # A regex that must match the path of the request for this override to be applied.
+    path_regex: re.Pattern
+
+    # The HTTP method of the request for the override to apply
+    verb: str
+
+    # The custom error class to use for this override.
+    error_class: type
+
+    # A regular expression that must match the error code for this override to be applied. If None,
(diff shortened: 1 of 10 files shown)


databricks/sdk/errors/mapper.py
```diff
@@ -1,8 +1,72 @@
...
+        if self.error_code_regex and not self.error_code_regex.match(raw_error.get('error_code', '')):
+            return False
+        if self.message_regex and not self.message_regex.match(raw_error.get('message', '')):
+            return False
+        return True
+
+
+_INVALID_PARAMETER_VALUE = 'INVALID_PARAMETER_VALUE'
+
+_all_overrides = [
```
nfx, line 46: @mgyucht why it is not consistent with https://github.com/databricks/databricks-sdk-go/pull/682? why is it not inhttps://github.com/databricks/databricks-sdk-go/blob/main/openapi/errors.go first? it'll be inconsistent among other languages otherwise.

nfx (changes requested):
it's not consistent with https://github.com/databricks/databricks-sdk-go/pull/682

databricks/sdk/errors/mapper.py
mgyucht: Addressed this now.

tanmay-db (approved):

PR #600, opened 2024-03-30
Check downstream backwards compatibility

nfx:
## Changes
Comments on PR whenever downstream fails to compile or fails unit tests. See implementation for more details:
- https://github.com/databrickslabs/sandbox/pull/141

Backwards incompatible changes would result in comments similar to:
image

Current downstream checking graph:

```mermaid
graph TD

pysdk[Python SDK: ***this PR***] --> ucx
pysdk --> blueprint

blueprint --> lsql
blueprint --> ucx
blueprint --> remorph

lsql --> ucx
lsql --> remorph
```

## Tests

Manual

### DIFF
--- a/.github/workflows/downstreams.yml
+++ b/.github/workflows/downstreams.yml
@@ -20,12 +20,14 @@ permissions:
   pull-requests: write
 
 jobs:
-  backwards-compatibility:
+  compatibility:
     strategy:
       fail-fast: false
       matrix:
         downstream:
           - name: ucx
+            org: databrickslabs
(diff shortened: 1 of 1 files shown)


nfx:
@mgyucht please also make the `ucx` BC-compat check as required.
image

tanmay-db (approved):
Change looks good to me but I don't think we should make it required as this would block SDK releases. 

Approving but please wait for @mgyucht to take a look as well as there were some concerns on having this enabled upstream instead.

PR #605, opened 2024-04-02
fix to set `filter` to `None` when `next_page_token` is returned from previous call 

kai-zhu-sonatype:
## Changes
clear `query[filter]` when preivous call returns `next_page_token` to avoid error when with `filter` request
```bash
databricks.sdk.errors.platform.InvalidParameterValue: When providing a page token and a filter, the filter must match the filter represented by the page token.
```
similar fix in https://github.com/databricks/databricks-sdk-java/pull/255, https://github.com/databricks/databricks-sdk-go/pull/875
## Tests

- [x] `make test` run locally
- [x] `make fmt` applied

### DIFF
--- a/databricks/sdk/service/pipelines.py
+++ b/databricks/sdk/service/pipelines.py
@@ -2001,6 +2001,7 @@ class PipelinesAPI:
             if 'next_page_token' not in json or not json['next_page_token']:
                 return
             query['page_token'] = json['next_page_token']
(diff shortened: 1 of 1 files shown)


kai-zhu-sonatype:
from the comment https://github.com/databricks/databricks-sdk-java/pull/255#issuecomment-2039369130, fixes will be on the server side

PR #618, opened 2024-04-17
Patch `dbutils.notebook.entry_point...` to return current local notebook path from env var

kartikgupta-db:
## Changes
* Users are using `dbutils.notebook.entry_point.getDbutils().notebook().getContext().notebookPath().get()` to get the current notebook path in notebooks (referring https://stackoverflow.com/questions/53523560/databricks-how-do-i-get-path-of-current-notebook). 
* This does not work from our current dbutils proxy. This PR patches this to read from a DATABRICKS_SOURCE_FILE env var when using local dbutils. 

## Proposal
* Allow users to add their own patches to make is easier for users using dbutils from sdk to workaround such issues in the future. 

## Tests
* Unit tests

- [x] `make test` run locally
- [x] `make fmt` applied


databricks/sdk/dbutils.py
```diff
@@ -241,6 +243,75 @@ def __getattr__(self, util) -> '_ProxyUtil':
...
+@dataclass
+class OverrideResult:
+    result: typing.Any
+
+
+def get_local_notebook_path():
+    value = os.getenv("DATABRICKS_SOURCE_FILE")
+    if value is None:
+        raise ValueError(
+            "DABRICKS_SOURCE_FILE environment variable is not set. This is required to get the local notebook path."
```
mgyucht, line 255:
```suggestion
            "Getting the current notebook path is only supported when running a notebook using the `Databricks Connect: Run as File` or `Databricks Connect: Debug as File` commands in the Databricks extension for VS Code. To bypass this error, set environment variable `DATABRICKS_SOURCE_FILE` to the desired notebook path."
```

databricks/sdk/dbutils.py
```diff
@@ -241,6 +243,75 @@ def __getattr__(self, util) -> '_ProxyUtil':
...
+    # This means, it is completely safe to override paths starting with `{util}.{attribute}.<other_parts>`, since none of the prefixes
+    # are being proxied to remote dbutils currently.
+    proxy_override_paths = {
+        'notebook.entry_point.getDbutils().notebook().getContext().notebookPath().get()':
+        get_local_notebook_path,
+    }
+
+    @classmethod
+    def __get_matching_overrides(cls, path: str):
+        return list(filter(lambda x: x.startswith(path), list(cls.proxy_override_paths.keys())))
```
mgyucht, line 284: List comprehension is slightly more readable?
```suggestion
        return [x for x in cls.proxy_override_paths.keys() if x.startswith(path)]
```

databricks/sdk/dbutils.py
```diff
@@ -241,6 +243,75 @@ def __getattr__(self, util) -> '_ProxyUtil':
...
+            return OverrideResult(self.proxy_override_paths[overrides[0]]())
+
+        if len(overrides) > 0:
+            return OverrideResult(_OverrideProxyUtil(name=path))
+
+        return None
+
+    def __call__(self, *args, **kwds) -> typing.Any:
+        if len(args) != 0 or len(kwds) != 0:
+            raise TypeError(f"{self._name}() takes no arguments (1 given)")
```
mgyucht, line 298: Suggestion: this is sort of faking an exception that Python would raise if a user called a function with extra arguments, but it might actually be confusing to users. Maybe it would be better to explicitly state something like `Arguments are not supported for overridden method {self._name}. Invoke as: {self._name}()`.

databricks/sdk/dbutils.py
```diff
@@ -241,6 +243,75 @@ def __getattr__(self, util) -> '_ProxyUtil':
                           name=util)
 
 
+@dataclass
+class OverrideResult:
```
mgyucht, line 247: Is this necessary? We just return `result.result` every time. Might as well return the result itself?

mgyucht (commented):
Few nits but otherwise looks good.

databricks/sdk/dbutils.py
kartikgupta-db: This is to provide a type that wraps a None result from the override. Some of the functions can then safely return None, if the override is not found or is not able to run.

mgyucht (approved):

PR #623, opened 2024-04-22
DBUtils implementation for Volumes

mgyucht:
## Changes
UC Volumes has been released for some time, but users are unable to use UC Volumes with `dbutils.fs` from the SDK. This PR implements support for `/Volumes` paths in DBUtils in the SDK.

I've done this primarily by extending `DbfsExt` to work with UC Volumes paths. A new class `_VolumePath` is supported, implementing the set of base operations defined on the `_Path` parent abstract class for volume paths. An accompanying `_VolumesIO` is also implemented to provide a consistent interface for reading from and writing to UC Volumes files, especially for writing: the `download` API already returns a `BinaryIO`, but the `upload` API _accepts_ a `BinaryIO`, so this adapter allows for a user to "open" a Volumes path for writing.

In order to properly implement `ls`, I changed the existing implementation of `list` in `_Path` subclasses to return a generator of `FileInfo`s. This allows for better reuse of this common functionality between `ls`, `cp` and `mv`.

Open questions:
* Should we do anything to distinguish between a path named `/Volumes` on disk and UC Volumes? E.g. via a different scheme?

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/databricks/sdk/mixins/files.py
+++ b/databricks/sdk/mixins/files.py
@@ -2,17 +2,16 @@ from __future__ import annotations
 
 import base64
 import functools
+import os
 import pathlib
 import shutil
 from abc import ABC, abstractmethod
 from io import BytesIO
 from types import TracebackType
-from typing import TYPE_CHECKING, AnyStr, BinaryIO, Iterable, Iterator, Type, Union, Generator
-import os
+from typing import (TYPE_CHECKING, AnyStr, BinaryIO, Generator, Iterable,
+                    Iterator, Type, Union)
 
-from databricks.sdk.core import DatabricksError
 from ..errors import NotFound
-
 from ..service import files
 
 if TYPE_CHECKING:
@@ -165,6 +164,7 @@ class _DbfsIO(BinaryIO):
 
 
 class _VolumesIO(BinaryIO):
+
     def __init__(self, api: files.FilesAPI, path: str, *, read: bool, write: bool, overwrite: bool):
         self._buffer = []
         self._api = api
@@ -323,6 +323,7 @@ class _Path(ABC):
 
 
 class _LocalPath(_Path):
+
     def _is_local(self) -> bool:
         return True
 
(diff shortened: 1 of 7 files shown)


databricks/sdk/mixins/files.py
```diff
@@ -238,69 +365,150 @@ def _list_local(self, recursive=False):
...
+            meta = self._api.get_metadata(self.as_string)
+            yield files.FileInfo(path=self.as_string,
+                                 is_dir=False,
+                                 file_size=meta.content_length,
+                                 modification_time=meta.last_modified,
+                                 )
+            return
+        queue = [self]
+        while queue:
+            next_path, queue = queue[0], queue[1:]
```
tanmay-db, line 437: Slicing the list will create a copy so this will be O(N), we can improve this to O(1) using `deque` and use `.popleft()` to get the first element.

databricks/sdk/mixins/files.py
```diff
@@ -238,69 +365,150 @@ def _list_local(self, recursive=False):
...
+            meta = self._api.get_status(self.as_string)
+            yield files.FileInfo(path='dbfs:' + self.as_string,
+                                 is_dir=False,
+                                 file_size=meta.file_size,
+                                 modification_time=meta.modification_time,
+                                 )
+            return
+        queue = [self]
+        while queue:
+            next_path, queue = queue[0], queue[1:]
```
tanmay-db, line 506: same here

databricks/sdk/mixins/files.py
```diff
@@ -333,67 +551,70 @@ def list(self, path: str, *, recursive=False) -> Iterator[files.FileInfo]:
...
+        p = self._path(path)
+        return p.exists()
 
     def _path(self, src):
         if str(src).startswith('file:'):
             return _LocalPath(src)
-        return _DbfsPath(self, src)
+        if src.startswith('dbfs:'):
+            src = src[len('dbfs:'):]
+        if str(src).startswith('/Volumes'):
```
tanmay-db, line 576: Should we consider typos here: https://docs.databricks.com/en/connect/unity-catalog/volumes.html

> Paths are also reserved for potential typos for these paths from Apache Spark APIs and dbutils, including /volumes, /Volume, /volume, whether or not they are preceded by dbfs:/. The path /dbfs/Volumes is also reserved, but cannot be used to access volumes.

tanmay-db (approved):

databricks/sdk/mixins/files.py
mgyucht: I think those paths are reserved but may not be usable themselves for looking up volumes (hopefully). Given that we use the REST API, we may need to be more narrow in what we accept than in DBR, since these paths correspond directly to REST API parameters.

PR #631, opened 2024-04-25
Remove unnecessary `dbfs:` scheme when listing in `w.dbutils.fs.ls()`

mgyucht:
## Changes
#623 introduced DBUtils support for volumes but also caused a small regression in listing behavior: `dbutils.fs.ls()` should not include the `dbfs:` scheme. This PR makes that fix. Additionally, it fixes a small bug in volumes recursive listing, only including the file paths as is the behavior with DBFS.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


tanmay-db (approved):

mgyucht:
Only failing integration tests are due to concurrent test runs: directories removed during recursive listing of WSFS cause the SDK to fail, and changes to the set of users while listing them causes the expected number of users to change while the test is running.

PR #632, opened 2024-04-29
Add 2 new TableTypes

sandormatyas:
## Changes

(rebased, expanded #476)

Adds two new values to TableType Enum:

- MANAGED_SHALLOW_CLONE - this is the value returned for shallow clone tables: https://docs.databricks.com/en/delta/clone-unity-catalog.html
- FOREIGN - this is the value returned for vector search indices: https://docs.databricks.com/en/generative-ai/create-query-vector-search.html

Unfortunately, these are not mentioned by the API documentation for the `table_type` field: https://docs.databricks.com/api/workspace/tables/list#tables-table_type

Here's an API call that fetches some tables, one with type MANAGED_SHALLOW_CLONE:
image

Here's an API call that fetches some tables, one with type FOREIGN:
image

Update:
Online tables also come with type FOREIGN, but different `data_source_format`:
image

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied


sandormatyas:
Hi @mgyucht, sorry for tagging you, but I was wondering if I could get some eyes on this or tips on how I could proceed?

We are fetching the table metadata using the sdk and we have multiple `None`-s now for table type.

renaudhartert-db:

PR #638, opened 2024-05-06
[Fix] Infer Azure tenant ID if not set

mgyucht:
## Changes
Port of https://github.com/databricks/databricks-sdk-go/pull/910 to the Python SDK.

In order to use Azure U2M or M2M authentication with the Databricks SDK, users must request a token from the Entra ID instance that the underlying workspace or account belongs to, as Databricks rejects requests to workspaces with a token from a different Entra ID tenant. However, with Azure CLI auth, it is possible that a user is logged into multiple tenants at the same time. Currently, the SDK uses the subscription ID from the configured Azure Resource ID for the workspace when issuing the `az account get-access-token` command. However, when users don't specify the resource ID, the SDK simply fetches a token for the active subscription for the user. If the active subscription is in a different tenant than the workspace, users will see an error such as:

```
io.jsonwebtoken.IncorrectClaimException: Expected iss claim to be: https://sts.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47/, but was: https://sts.windows.net/e3fe3f22-4b98-4c04-82cc-d8817d1b17da/
```

This PR modifies Azure CLI and Azure SP credential providers to attempt to load the tenant ID of the workspace if not provided before authenticating. Currently, there are no unauthenticated endpoints that the tenant ID can be directly fetched from. However, the tenant ID is indirectly exposed via the redirect URL used when logging into a workspace. In this PR, we fetch the tenant ID from this endpoint and configure it if not already set.

Here, we lazily fetch the tenant ID only in the auth methods that need it. This prevents us from making any unnecessary requests if these Azure credential providers are not needed.

## Tests
Unit tests check that the tenant ID is fetched automatically if not specified for an azure workspace when authenticating with client ID/secret or with the CLI.

- [x] `make test` run locally
- [x] `make fmt` applied


databricks/sdk/config.py
```diff
@@ -363,6 +363,31 @@ def _fix_host_if_needed(self):
...
+        entra_id_endpoint = resp.headers.get('Location')
+        if entra_id_endpoint is None:
+            logger.debug(f'No Location header in response from {login_url}')
+            return
+        url = urllib.parse.urlparse(entra_id_endpoint)
+        path_segments = url.path.split('/')
+        if len(path_segments) < 2:
+            logger.debug(f'Invalid path in Location header: {url.path}')
+            return
+        self.azure_tenant_id = path_segments[1]
```
hectorcast-db, line 388: Nit, Can we have a sample of the URL as a documentation?

tests/test_config.py
```diff
@@ -74,3 +77,40 @@ def test_config_copy_deep_copies_user_agent_other_info(config):
     assert "blueprint/0.4.6" in config.user_agent
     assert "blueprint/0.4.6" in config_copy.user_agent
     useragent._reset_extra(original_extra)
+
+
+def test_load_azure_tenant_id_404(requests_mock, monkeypatch):
+    monkeypatch.setenv('PATH', __tests__ + '/testdata:/bin')
```
hectorcast-db, line 83: It seems to be causing issues with Windows tests

hectorcast-db (commented):
It seems Windows tests are failing

hectorcast-db (approved):

PR #641, opened 2024-05-10
Gracefully handle null JSON responses from API

sodle-splunk:
## Changes

For some DELETE requests, notably when deleting an Alert, the API returns a JSON response containing only `null`. This is not seen as an empty response, because the content length is greater than zero.

When calling `WorkspaceClient.alerts.delete(alertId)`, this results in the alert being deleted as expected. However, a `TypeError` is raised when trying to unpack the response at `databricks/sdk/core.py:170`.

To fix this, we will null-check the decoded JSON before attempting to unpack it. If it is `None`, we will simply return the response dict like we do for empty responses.

## Tests

- [X] `make test` run locally
- [X] `make fmt` applied
  - This seems to have modified far more lines than expected.
  - Actual code changes are [here](https://github.com/databricks/databricks-sdk-py/pull/641/files#diff-62999ba950664150f7db28196ce10b0ef471f324264425a682a169b109ce5c03R153)
  - New test is [here](https://github.com/databricks/databricks-sdk-py/pull/641/files#diff-938de46e643ab091cff6a7c23e4e752e8907e2266f47889d988923352f7a1058R344)
- [X] relevant integration tests applied
  - Added a test which shows the `null` response received when calling `DELETE /api/2.0/preview/sql/alerts/alertid`. This test fails before applying my fix, and passes with the fix in place.

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -146,6 +146,8 @@ class ApiClient:
             return resp
 
         json = response.json()
+        if json is None:
+            return resp
         if isinstance(json, list):
             return json
 
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -14,9 +14,9 @@ from .credentials_provider import *
 from .errors import DatabricksError, error_mapper
 from .retries import retried
 
-__all__ = ['Config', 'DatabricksError']
+__all__ = ["Config", "DatabricksError"]
 
-logger = logging.getLogger('databricks.sdk')
+logger = logging.getLogger("databricks.sdk")
 
 
 class ApiClient:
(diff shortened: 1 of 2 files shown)


sodle (approved):

mgyucht:
I believe @tanmay-db was working on a fix for this case: https://github.com/databricks/databricks-sdk-py/pull/579.

tanmay-db:
Hi @sodle-splunk, thanks for the PR. The formatting changes might be due to different versions of yapf being used, I ran make fmt on main branch and there were no diffs. 

The test looks good and I tried to revert the make fmt commit so we could merge this PR but couldn't since I don't have the permission to push to your fork:
````
To github.com:sodle-splunk/databricks-sdk-py.git
 ! [remote rejected]   head -> sodle-splunk/main (permission denied)
error: failed to push some refs to 'github.com:sodle-splunk/databricks-sdk-py.git'
````

I have cherry picked the commits and added them to the PR: https://github.com/databricks/databricks-sdk-py/pull/579

sodle-splunk:
Ah, I could have sworn I'd allowed maintainer commits on this PR. Oh well.

Thanks for cherry-picking. I'll close this one.

PR #644, opened 2024-05-16
Create a method to generate OAuth tokens

hectorcast-db:
## Changes
Add method to get OAuth tokens

## Tests
- [X] `make test` run locally
- [X] `make fmt` applied
- [X] Manual test (cannot be run as integration tests due to limitations in the current infrastructure setup)
```
def test():
    w = WorkspaceClient(profile='DEFAULT')
    auth_details = f'"type":"workspace_permission","object_type":"serving-endpoints","object_path":"/serving-endpoints/REDACTED","actions":["query_inference_endpoint"]'
    auth_details = "[{" + auth_details + "}]"
    t = w.api_client.get_oauth_token(auth_details)
    print(t)
```
Result:
```
Token(access_token='REDACTED', token_type='Bearer', refresh_token=None, expiry=datetime.datetime(2024, 5, 16, 11, 9, 8, 221008))
```


databricks/sdk/core.py
```diff
@@ -108,6 +109,21 @@ def flatten_dict(d: Dict[str, Any]) -> Dict[str, Any]:
         flattened = dict(flatten_dict(with_fixed_bools))
         return flattened
 
+    def get_oauth_token(self, auth_details: str) -> Token:
```
mgyucht, line 112: Does it make sense for some of this code to live in `oauth.py`?

databricks/sdk/core.py
```diff
@@ -108,6 +109,21 @@ def flatten_dict(d: Dict[str, Any]) -> Dict[str, Any]:
...
+        if not self._cfg.auth_type:
+            self._cfg.authenticate()
+        original_token = self._cfg.oauth_token()
+        path = "/oidc/v1/token"
+        headers = {"Content-Type": "application/x-www-form-urlencoded"}
```
mgyucht, line 117: Let's move these constants somewhere that they can be consistently referenced.

databricks/sdk/credentials_provider.py
```diff
@@ -27,6 +27,17 @@
...
+    @abc.abstractmethod
+    def __call__(self) -> Dict[str, str]:
+        ...
+
+    @abc.abstractmethod
+    def oauth_token(self) -> Token:
+        ...
+
+
 class CredentialsProvider(abc.ABC):
```
mgyucht, line 41: Shall we rename CredentialsProvider to CredentialsStrategy here as well?

databricks/sdk/credentials_provider.py
```diff
@@ -49,12 +69,34 @@ def inner(func: Callable[['Config'], HeaderFactory]) -> CredentialsProvider:
...
+
+        @functools.wraps(func)
+        def wrapper(cfg: 'Config') -> Optional[OAuthHeaderFactory]:
             for attr in require:
                 if not getattr(cfg, attr):
                     return None
             return func(cfg)
 
         wrapper.auth_type = lambda: name
+        wrapper.oauth_token = lambda cfg: wrapper(cfg).oauth_token()
```
mgyucht, line 99: This is kind of strange looking, but I think it does work. We've defined OauthCredentialsProvider as a metaclass, but maybe we could define an instance of it here. I'm worried because I'm not sure the Python type checking is powerful enough to verify that we're setting the correct attribute on the inner classes. Better might be to define OauthCredentialsProvider as a proper class with an `__init__` method, where these functions can be passed in via constructor. That way, we get loud failures if we forget to specify some element of the credential provider.

databricks/sdk/credentials_provider.py
```diff
@@ -277,6 +323,10 @@ def google_credentials(cfg: 'Config') -> Optional[HeaderFactory]:
     gcp_credentials = service_account.Credentials.from_service_account_info(info=account_info,
                                                                             scopes=GcpScopes)
 
+    def token() -> Token:
+        credentials.refresh(request)
```
mgyucht, line 327: Do these need to be refreshed every time?

databricks/sdk/credentials_provider.py
```diff
@@ -601,6 +658,19 @@ def __init__(self) -> None:
     def auth_type(self) -> str:
         return self._auth_type
 
+    def oauth_token(self, cfg: 'Config') -> Token:
+        auth_providers = [
+            pat_auth, basic_auth, metadata_service, oauth_service_principal, azure_service_principal,
+            github_oidc_azure, azure_cli, external_browser, databricks_cli, runtime_native_auth,
+            google_credentials, google_id
+        ]
```
mgyucht, line 666: Let's just initialize this in the `__init__` constructor rather than copying it.

mgyucht (commented):
Some suggestions, but coming along nicely

databricks/sdk/core.py
hectorcast-db: I did a few changes to reuse some function from `oauth.py`. However, this method still makes sense for 2 reasons:
* Config is not exposed int he cli and we need it
* Even after exposing the config, we can't pass it to `oauth.py` or we create circular dependencies.
So having this here is helpful because it simplifies things. I could move the creation of `params` inside `oauth.py`.

databricks/sdk/credentials_provider.py
hectorcast-db: Note that we do the same for normal credentials.

But in this case, I think it makes even more sense. We are going to cache the resulting Token, and the resulting Token has the same expiration as the original one. So refreshing it on creation makes sense so it will last the longest possible.

.codegen/__init__.py.tmpl
```diff
@@ -46,11 +46,11 @@ class WorkspaceClient:
                  debug_headers: bool = None,
                  product="unknown",
                  product_version="0.0.0",
-                 credentials_provider: CredentialsProvider = None,
```
mgyucht, line 49: Let's respect credentials_provider here as well, like what you do in Config

databricks/sdk/config.py
```diff
@@ -82,15 +81,19 @@ class Config:
...
                  product="unknown",
                  product_version="0.0.0",
                  clock: Clock = None,
                  **kwargs):
         self._header_factory = None
         self._inner = {}
         self._user_agent_other_info = []
-        self._credentials_provider = credentials_provider if credentials_provider else DefaultCredentials()
+        if credentials_strategy and credentials_provider:
+            raise ValueError("When providing `credentials_strategy` field, `credential_provider` cannot be specified.")
```
mgyucht, line 95: Suggestion: let's log a warning here if credentials_provider attribute is set.

mgyucht (approved):
Two small remarks, otherwise this looks good! Feel free to click the merge button when ready.

JCZuurmond:
@hectorcast-db : Something looks off with the type hints: [this](https://github.com/databricks/databricks-sdk-py/blob/f7d920e1f912b204669b826ed76e026607c59797/databricks/sdk/credentials_provider.py#L116) does not match type returned.

Had to ignore the types in https://github.com/databrickslabs/ucx/pull/1943

How should this be handled?

PR #655, opened 2024-05-24
Revert "Revert "Create a method to generate OAuth tokens (#644)" (#653)"

hectorcast-db:
This reverts commit 9c5fae7eae1e84cdc8b635541610e7e8114b0615.

## Changes
Introduce again the method to generate OAuth tokens, including fixes for impacted auth types

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied

### DIFF
--- a/databricks/sdk/credentials_provider.py
+++ b/databricks/sdk/credentials_provider.py
@@ -263,6 +263,7 @@ def azure_service_principal(cfg: 'Config') -> CredentialsProvider:
 
     return OAuthCredentialsProvider(refreshed_headers, token)
 
+
 @oauth_credentials_strategy('github-oidc-azure', ['host', 'azure_client_id'])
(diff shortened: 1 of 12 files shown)


databricks/sdk/credentials_provider.py
```diff
@@ -124,11 +176,14 @@ def inner() -> Dict[str, str]:
         token = token_source.token()
         return {'Authorization': f'{token.token_type} {token.access_token}'}
 
-    return inner
+    def token() -> Token:
```
hectorcast-db, line 179: The difference from the original is this function. Previous code:
```
return OAuthCredentialsProvider(inner, inner.token)
```
That does not work because `config.py` tries to make a copy of `inner` which uses a "lock" (and cannot be copied). The function works as a wrapper, and allows this object to be copied without copying `inner`.

tanmay-db (approved):

PR #666, opened 2024-06-03
adding compute field to the Config

alexkh-db:
## Changes
We are currently adding serverless support to Databricks Connect client. Serverless can be enabled in the client using `DatabricksSession.builder.serverless()`. We also want to add support for serverless in the databricks config.

- We propose adding a new `compute` attribute in the config that will be new unifying field that can contain values `"serverless"` or `"cluster:<cluster_id>"`. `cluster_id` will be deprecated but still supported after this change. We could add support for `warehouse_id` in the `compute` field as well if necessary.
- `compute` and `cluster_id` should be mutually exclusive, if both are specified as kwargs, env variables or fields in `.databrickscfg`, an error will be thrown.
- `get_compute` method returns a `Compute` object that parses `.compute` and `.cluster_id` fields and provides convinient `is_empty()`, `is_serverless()`, and `get_cluster_id()` methods.

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied

### DIFF
--- a/databricks/sdk/oauth.py
+++ b/databricks/sdk/oauth.py
@@ -371,9 +371,6 @@ class OAuthClient:
         config = Config(host=host, credentials_strategy=noop_credentials)
         if not scopes:
             scopes = ['all-apis']
-        if config.is_azure:
-            # Azure AD only supports full access to Azure Databricks.
-            scopes = [f'{config.effective_azure_login_app_id}/user_impersonation', 'offline_access']
         oidc = config.oidc_endpoints
         if not oidc:
             raise ValueError(f'{host} does not support OAuth')
@@ -385,6 +382,7 @@ class OAuthClient:
         self.token_url = oidc.token_endpoint
         self.is_aws = config.is_aws
         self.is_azure = config.is_azure
+        self.is_gcp = config.is_gcp
 
         self._auth_url = oidc.authorization_endpoint
         self._scopes = scopes
--- a/docs/oauth.md
+++ b/docs/oauth.md
@@ -245,8 +245,6 @@ Databricks SDK for Python exposes the `oauth_client.initiate_consent()` helper t
 PKCE state verification. Application developers are expected to persist `RefreshableCredentials` in the webapp session
 and restore it via `RefreshableCredentials.from_dict(oauth_client, session['creds'])` helpers.
 
-Works for both AWS and Azure. Not supported for GCP at the moment.
-
 ```python
 from databricks.sdk.oauth import OAuthClient
 oauth_client = OAuthClient(host='<workspace-url>',
@@ -305,10 +303,6 @@ account_client = AccountClient(host='https://accounts.cloud.databricks.com',
                                account_id=input('Databricks Account ID: '),
                                username=input('Username: '),
                                password=getpass.getpass('Password: '))
-logging.info('Enrolling all published apps...')
-account_client.o_auth_enrollment.create(enable_all_published_apps=True)
-status = account_client.o_auth_enrollment.get()
-logging.info(f'Enrolled all published apps: {status}')
 custom_app = account_client.custom_app_integration.create(
     name='awesome-app',
     redirect_urls=[f'https://host.domain/path/to/callback'],
--- a/examples/flask_app_with_oauth.py
+++ b/examples/flask_app_with_oauth.py
@@ -8,14 +8,14 @@ access Databricks resources.
 
 If you have already Custom App:
 
-./flask_app_with_oauth.py <databricks workspace url> \
+./flask_app_with_oauth.py --host <databricks workspace url> \
     --client_id <app-client-id> \
     --client_secret <app-secret> \
     --port 5001
 
 If you want this script to register Custom App and redirect URL for you:
 
-./flask_app_with_oauth.py <databricks workspace url>
+./flask_app_with_oauth.py --port 5001 --profile <databricks account profile>
 
 You'll get prompted for Databricks Account username and password for
 script to enroll your account into OAuth and create a custom app with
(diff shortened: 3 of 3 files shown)


databricks/sdk/config.py
```diff
@@ -425,13 +438,25 @@ def _known_file_config_loader(self):
...
+
+        if self.compute is None and self.cluster_id is None:
+            if "compute" in raw_config and "cluster_id" in raw_config:
+                raise ValueError("Cannot specify both compute and cluster id in the config file")
         for k, v in raw_config.items():
             if k in self._inner:
                 # don't overwrite a value previously set
                 continue
+            if (k == "compute" and self.cluster_id is not None
+                    or k == "cluster_id" and self.compute is not None):
```
nija-at, line 450: Add some parenthesis so it's easier to read which operators are grouped.

```suggestion
            if (k == "compute" and self.cluster_id is not None)
                    or (k == "cluster_id" and self.compute is not None):
```

databricks/sdk/config.py
```diff
@@ -473,3 +498,36 @@ def deep_copy(self):
...
+        elif cfg.compute is not None:
+            if cfg.compute == "serverless":
+                self.serverless = True
+            elif cfg.compute.startswith("cluster:"):
+                self.cluster_id = cfg.compute.split("cluster:")[1]
+            else:
+                raise ValueError("compute must be either 'serverless' or 'cluster:<cluster_id>'")
+        elif cfg.cluster_id is not None:
+            self.cluster_id = cfg.cluster_id
+
```
nija-at, line 522: Looks like this is cross-populating values, i.e., should Compute be populated by the value of `DATABRICKS_CLUSTER_ID`? 
I thought we decided it would not, right?

databricks/sdk/config.py
```diff
@@ -378,12 +381,22 @@ def _set_inner_config(self, keyword_args: Dict[str, any]):
...
+                                 "DATABRICKS_CLUSTER_ID in the environment.")
+
         found = False
         for attr in self.attributes():
             if not attr.env:
                 continue
             if attr.name in self._inner:
                 continue
+            if (attr.name == "compute" and self.cluster_id is not None
+                    or attr.name == "cluster_id" and self.compute is not None):
```
nija-at, line 397: same about paranthesis and grouping

databricks/sdk/config.py
```diff
@@ -473,3 +498,36 @@ def deep_copy(self):
...
+                self.serverless = True
+            elif cfg.compute.startswith("cluster:"):
+                self.cluster_id = cfg.compute.split("cluster:")[1]
+            else:
+                raise ValueError("compute must be either 'serverless' or 'cluster:<cluster_id>'")
+        elif cfg.cluster_id is not None:
+            self.cluster_id = cfg.cluster_id
+
+    def is_empty(self):
+        return self.serverless is False and self.cluster_id is None
```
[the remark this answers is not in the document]
alexkh-db: i thought this would be convenient. The downstream code will need to either check
```
if config.compute is None and config.cluster_id is None:
```
or
```
compute = config.get_compute()
if compute.serverless is False and compute.cluster_id is None:
```
which is longer than using a single method

databricks/sdk/config.py
alexkh-db: it's not really cross-population, `Compute` is an adapter object to make it easier to get `serverless` and `cluster_id` values from the raw `.compute` and `.cluster_id` attributes. `config.compute` and `config.cluster_id` are not changed and remain the same as specified by the user

databricks/sdk/config.py
```diff
@@ -378,12 +381,22 @@ def _set_inner_config(self, keyword_args: Dict[str, any]):
...
+                raise ValueError("Cannot specify both DATABRICKS_COMPUTE and "
+                                 "DATABRICKS_CLUSTER_ID in the environment.")
+
         found = False
         for attr in self.attributes():
             if not attr.env:
                 continue
             if attr.name in self._inner:
                 continue
+            if ((attr.name == "compute" and self.cluster_id is not None)
```
mgyucht, line 396: I think this is going to be confusing as users may not be 100% familiar with the precedence between different options. Would it be better to require this to be unambiguous, i.e. either cluster_id or compute is set at exactly one place?

databricks/sdk/config.py
```diff
@@ -473,3 +498,36 @@ def deep_copy(self):
...
+    def get_compute(self):
+        return Compute(self)
+
+
+class Compute:
+    serverless: bool = False
+    cluster_id: Optional[str] = None
+
+    def __init__(self, cfg: Config):
+        if cfg.compute is not None and cfg.cluster_id is not None:
```
mgyucht, line 511: I think this is duplicated logic from the Config `__init__` and descendants?

databricks/sdk/config.py
```diff
@@ -473,3 +498,36 @@ def deep_copy(self):
         """Creates a deep copy of the config object.
         """
         return copy.deepcopy(self)
+
+    def get_compute(self):
+        return Compute(self)
+
+
+class Compute:
```
mgyucht, line 506: Is this class really necessary? Is there a downside to attaching the `is_serverless()` and `get_cluster_id()` methods to `Config` directly? Then all of lines 501-533 become
```python
  @property
  def is_serverless(self):
    return self.compute == "serverless"

  @property 
  def get_cluster_id(self):
    if self.compute.startswith("cluster:"):
      return self.compute[len("cluster:"):]
    return self.cluster_id
```

mgyucht (commented):
Some suggestions to clean this up. I think we should try to keep logic from spreading to different parts of the init procedure if possible, so I have some suggestions on how to move validation to the `_validate` method.

Issue #671, opened 2024-06-10
Using Azure App Registration with 'Mobile and desktop applications' fails refresh because of Origin header


thijs-nijhuis:
**Description**
We have an App Registration on Azure which used the 'Mobile and desktop applications' platform authentication. Getting a new token using the authorization code flows works as expected. But when a refresh of the token is needed, it fails with the following error
```
AADSTS9002326: Cross-origin token redemption is permitted only for the 'Single-Page Application' client-type. Request origin: 'http://localhost:8020'.
```
I think this is caused by the fact that the header is always [set](https://github.com/databricks/databricks-sdk-py/blob/455a14c1e34504f7da24c105ccdca9812b5c07db/databricks/sdk/oauth.py#L258). Also for non-spa app registrations.

When getting a new token from scratch, a try/except strategy is [used](https://github.com/databricks/databricks-sdk-py/blob/455a14c1e34504f7da24c105ccdca9812b5c07db/databricks/sdk/oauth.py#L317) to make it work for both spa and non-spa platform. I think this would work here as well and also be more consistent.

**Reproduction**
I use dbt-databricks in which I encountered [this error](https://github.com/databricks/dbt-databricks/issues/640#issuecomment-2155071792). Therefor, I do not have any stand alone code.

**Expected behavior**
Be able to refresh token for both SPA and 'Mobile and desktop applications' platforms.

**Is it a regression?**
I don't think so. 

**Debug Logs**
N.A.

**Other Information**
 - OS: Windows
 - Version: 11 enterprise

**Additional context**
I will try to solve and test this and create a PR linking to this issue.

thijs-nijhuis:
Hi, can anyone please comment on this issue and perhaps trigger the test on the corresponding PR?

PR #672, opened 2024-06-10
Added retry logic to token refresh to deal with

thijs-nijhuis:
## Changes
The logic that fetches an initial access_token has a fall back/retry path for SPA applications which need an 'origin' header. For the part that fetches a refresh token, this fall back was not there. Instead, an 'origin' would always be set for Azure endpoints but this causes issues with application that use the 'Mobile and desktop applications' platform authentication which will throw an error if an 'origin' header is set. This change should fix [this](https://github.com/databricks/databricks-sdk-py/issues/671) issue by tackling both paths in the same way.

## Tests
- [x] `make test` run locally
- [x] `make fmt` applied


sertacosh:
We need this change implemented!

renaudhartert-db:

Issue #660, opened 2024-05-28
[ISSUE] InvalidParameterValue: Path must be absolute: \mnt


ridvansg:
**Description**
If I try to check if a dbfs folder exists on databricks instance, with code like client.dbfs.exists('/mnt'), I get error "InvalidParameterValue: Path must be absolute: \mnt". 

The issue happens on windows only.

**Reproduction**
1. On Windows start ipython
2. use following code:
`import logging
 import argparse
 from databricks.sdk import WorkspaceClient
 from databricks.sdk.service.compute import DataSecurityMode, RuntimeEngine, Library
 from datetime import timedelta
 import os
 client = WorkspaceClient()
 client.dbfs.exists('/mnt')`
3. Error message:
`InvalidParameterValue: Path must be absolute: \mnt`

**Expected behavior**
"true" or "false", depending on whether the path exists on databricks.

**Is it a regression?**
I didn't try.

**Debug Logs**
'Cell In[11], line 1
----> 1 client.dbfs.exists('/mnt')

File C:\ProgramData\anaconda3\envs\databricks2\lib\site-packages\databricks\sdk\mixins\files.py:572, in DbfsExt.exists(self, path)
    570 """If file exists on DBFS"""
    571 p = self._path(path)
--> 572 return p.exists()

File C:\ProgramData\anaconda3\envs\databricks2\lib\site-packages\databricks\sdk\mixins\files.py:490, in _DbfsPath.exists(self)
    488 def exists(self) -> bool:
    489     try:
--> 490         self._api.get_status(self.as_string)
    491         return True
    492     except NotFound:

File C:\ProgramData\anaconda3\envs\databricks2\lib\site-packages\databricks\sdk\service\files.py:624, in DbfsAPI.get_status(self, path)
    621 if path is not None: query['path'] = path
    622 headers = {'Accept': 'application/json', }
--> 624 res = self._api.do('GET', '/api/2.0/dbfs/get-status', query=query, headers=headers)
    625 return FileInfo.from_dict(res)

File C:\ProgramData\anaconda3\envs\databricks2\lib\site-packages\databricks\sdk\core.py:132, in ApiClient.do(self, method, path, query, headers, body, raw, files, data, response_headers)
    128 headers['User-Agent'] = self._user_agent_base
    129 retryable = retried(timeout=timedelta(seconds=self._retry_timeout_seconds),
    130                     is_retryable=self._is_retryable,
    131                     clock=self._cfg.clock)
[...]

PR #673, opened 2024-06-11
fix windows path (#660)

parthban-db:
## Changes

Changed `pathlib.Path` with the `pathlib.PurePosixPath` in `/databricks/sdk/mixins/files.py` which always use linux path separators regardless of the OS that it is running on. Fixes (#660)

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied


databricks/sdk/mixins/files.py
```diff
@@ -267,7 +267,8 @@ def __repr__(self) -> str:
 class _Path(ABC):
 
     def __init__(self, path: str):
-        self._path = pathlib.Path(str(path).replace('dbfs:', '').replace('file:', ''))
+        self._path = pathlib.PurePosixPath(str(path).replace('dbfs:', '').replace('file:', ''))
+        self._system_path = pathlib.Path(self._path)
```
edwardfeng-db, line 271: It'd be good to add a comment explaining why we need `self._path` to be `PurePosixPath` and why we need `self._system_path`

edwardfeng-db:
Can you link the github issue that this PR solves in the PR description?

edwardfeng-db (approved):
LGTM, let's get @mgyucht to take a look before merging

mgyucht (changes requested):
Thank you for your contribution @parthban-db! A couple notes about this PR:

The more I think about this, the more I think we're putting too much implementation detail in one place. I would do the following:
1. Change `_Path.__init__` to be an `@abstractmethod` and remove its implementation.
2. Add `__init__` implementations in each of `_DbfsPath`, `_LocalPath` and `_VolumesPath`. In `_DbfsPath` and `_VolumesPath`, use `PurePosixPath`, and in `_LocalPath` use `Path`.
This way, we don't duplicate the `path` state, and each implementation can choose the specific path type needed for itself.

Second: We should also make sure we add a test case that covers the actual issue reported by the user: when making a directory in DBFS, regardless of the OS, the path separator used in the API call must be `/`, not `\`. 

@edwardfeng-db could you help @parthban-db modify the unit test runners to support testing in Windows as well?

tests/test_dbfs_mixins.py
```diff
@@ -70,3 +70,22 @@ def test_fs_path_invalid(config):
...
+
+def test_dbfs_local_path_mkdir(config, tmp_path):
+    from databricks.sdk import WorkspaceClient
+
+    w = WorkspaceClient(config=config)
+    w.dbfs._path(f'file:{tmp_path}/test_dir').mkdir()
+    assert w.dbfs.exists(f'file:{tmp_path}/test_dir')
+
+
+def test_dbfs_exists(config, mocker):
```
mgyucht, line 83: Can we do this test for volumes as well?

mgyucht (commented):
LGTM once you add the volumes test

mgyucht (approved):

PR #675, opened 2024-06-11
Add `serverless_compute_id` field to the config

alexkh-db:
## Changes
We propose adding a new `serverless_compute_id` attribute in the config that will be used to enable serverless compute in the downstream applications. The default value to enable serverless is `serverless_compute_id = auto`, other values might be used in the future if needed. Config doesn't have to validate the value in `serverless_compute_id`, this will be done in the downstream applications.

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied


nija-at (commented):
LGTM. Docs?

alexkh-db:
> LGTM. Docs?

I couldn't find any docs for the other related fields, e.g. cluster_id, so i think we are not documenting it (at least yet). @mgyucht please lmk if documentation is needed

alexkh-db:
> > Config doesn't have to validate the value in serverless_gc_id, this will be done in the downstream applications.
> 
> Isn't the only allowed value at the moment "auto" then? If this is the case, shouldn't it have validation?

i'm open to both options, the pros for not validating in config IMO are:
1) config is doing little validation for other fields and mostly leaves validation to the downstream applications. not sure if we should single out `serverless_gc_id` field
2) if dbconnect or any other downstream applications will need to add another value except `auto`, databricks-sdk code doesn't need to change and the user has one less package to upgrade

grundprinzip:
Sorry for being horrible annoying. According to https://docs.databricks.com/en/compute/serverless.html I would just name the field `serverless_compute_id`.

databricks/sdk/config.py
```diff
@@ -66,6 +66,7 @@ class Config:
     auth_type: str = ConfigAttribute(env='DATABRICKS_AUTH_TYPE')
     cluster_id: str = ConfigAttribute(env='DATABRICKS_CLUSTER_ID')
     warehouse_id: str = ConfigAttribute(env='DATABRICKS_WAREHOUSE_ID')
+    serverless_compute_id: str = ConfigAttribute(env='SERVERLESS_COMPUTE_ID')
```
grundprinzip, line 69:
```suggestion
    serverless_compute_id: str = ConfigAttribute(env='DATABRICKS_SERVERLESS_COMPUTE_ID')
```

grundprinzip: Sorry, found one more thing.

alexkh-db: ah, yes, makes sense

alexkh-db: my bad :)

grundprinzip (approved):

pietern (approved):

Issue #661, opened 2024-05-28
[ISSUE] Incorrectly set config.is_aws flag when host has trailing "/"


HariGS-DB:
**Description**
When running databricks sdk on azure workspaces using databricks cli authentication. The config.is_aws flag is set to True.
This occurs when the databricks profile host doesn't have https://  in the beginning and a trailing "/" at the end for a azure workspace

**Reproduction**
Create a databricks profile like below:
azure workspace url
host = adb-<workspace url>/
auth_type = databricks-cli

run code :
ws = WorkspaceClient(auth_type = "databricks-cli", profile="PROFILE_NAME")
assert ws.config.is_azure

**Expected behaviour**
ws.config.is_azure should be set to True and ws.config.is_aws should be set to False

**Is it a regression?**
Dont know

**Other Information**
 - OS: [e.g. macOS]
 - Version: [e.g. 0.28.0]

**Additional context**
related ucx issue
https://github.com/databrickslabs/ucx/issues/1757

HariGS-DB:
confirmed working and closing.

PR #681, opened 2024-06-17
Check trailing slash in host url

parthban-db:
## Changes

Added a check for trailing slash in the host url. Fixes (#661)

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


tests/test_config.py
```diff
@@ -15,3 +15,21 @@ def test_config_supports_legacy_credentials_provider():
     c2 = c.copy()
     assert c2._product == 'foo'
     assert c2._product_version == '1.2.3'
+
+
+def test_config_host_url_format_check(mocker):
```
mgyucht, line 20: Take a look at https://docs.pytest.org/en/7.0.x/how-to/parametrize.html

mgyucht (commented):
One suggestion on the test, but otherwise LGTM

mgyucht (approved):

PR #684, opened 2024-06-19
Update jobs.py: fix invalid escape sequence (issue #508)

### LINKED ISSUE #508: [ISSUE]  DeprecationWarning: invalid escape sequence '\.'
**Description**
https://github.com/databricks/databricks-sdk-py/blob/a558d97415e8ced4b88c70e03e0a31f4f2006610/databricks/sdk/service/jobs.py#L61C30-L61C30 the backslash before the period in the docstring throws the following warning. DeprecationWarning: invalid escape sequence '\.'.

**Reproduction**
Run the following code through pytest or something. 
```py
"""The sequence number of this run attempt for a triggered job run. The initial attempt of a run
    has an attempt_number of 0\. If the initial run attempt fails, and the job has a retry policy
    (`max_retries` \> 0), subsequent runs are created with an `original_attempt_run_id` of the
    original attempt’s ID and an incrementing `attempt_number`. Runs are retried only until they
    succeed, and the maximum `attempt_number` is the same as the `max_retries` value for the job."""
```

**Expected behaviour**
Not receiving a warning.

**Is it a regression?**
Didn't check.

**Other Information**
 - OS: Linux.
 - Python: 3.10.
 - Version: latest.

wyattscarpenter:
## Changes
Fixes #508, by making several offending strings into raw strings.

## Tests
I used `uv run ruff check --select W605  . --no-respect-gitignore` on the directory where I have the last distribution of the code, and it seems to think these are the only invalid escape sequences in the project.


wyattscarpenter:
Today I was reminded by some warnings in my terminal that this is still a live issue. I would be pleased if it were merged, although it's ultimately of little importantance. 

I will update this branch to match the base to keep this pr relevant, as GitHub seems to recommend.

wyattscarpenter:
I think this was probably fixed in another way (removing the backslashes entirely), in Sugust, with https://github.com/databricks/databricks-sdk-py/commit/fb30ed98912a4e62a4b535188765f792b7bc6f47#diff-9125abd285f0b94b52f0445f9f85a75a921fae47124d670ab9f8f822de9f33acL6
Hence this pr now has a merge conflict. And, more importantly, is redundant.

PR #697, opened 2024-07-05
Fix auth tests for windows.

parthban-db:
## Changes

Fix auth tests for windows. 

- Added a powershell script as bash script doesn't run on windows
- change 'COMSPEC' enviornment variable to run commands on powershell
- Use 'USERPROFILE' instead of 'HOME' as it is alternative of 'HOME' in windows

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -39,7 +39,11 @@ def raises(msg):
             with pytest.raises(ValueError) as info:
                 func(*args, **kwargs)
             exception_str = str(info.value)
-            exception_str = exception_str.replace(__tests__ + '/', '')
+            if platform.system() == 'Windows':
+                exception_str = exception_str.replace(__tests__ + '\\', '')
+                exception_str = exception_str.replace('\\', '/')
+            else:
+                exception_str = exception_str.replace(__tests__ + '/', '')
             assert msg in exception_str
 
         return wrapper
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,6 @@
 import functools
 import os
+import platform
 
 import pytest as pytest
 from pyfakefs.fake_filesystem_unittest import Patcher
@@ -57,3 +58,18 @@ def fake_fs():
         patcher.fs.add_real_directory(test_data_path)
 
         yield patcher.fs # This will return a fake file system
+
+
+def set_home(monkeypatch, path):
+    if platform.system() == 'Windows':
+        monkeypatch.setenv('USERPROFILE', __tests__ + path)
+    else:
+        monkeypatch.setenv('HOME', __tests__ + path)
+
+
+def set_az_path(monkeypatch):
+    if platform.system() == 'Windows':
+        monkeypatch.setenv('Path', __tests__ + "\\testdata\\windows\\")
+        monkeypatch.setenv('COMSPEC', 'C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe')
+    else:
+        monkeypatch.setenv('PATH', __tests__ + "/testdata:/bin")
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -121,7 +121,7 @@ def test_config_config_file(monkeypatch):
 
 @raises(f"{default_auth_base_error_message}. Config: host=https://x")
 def test_config_config_file_skip_default_profile_if_host_specified(monkeypatch):
-    monkeypatch.setenv('HOME', __tests__ + '/testdata')
+    set_home(monkeypatch, '/testdata')
     cfg = Config(host='x')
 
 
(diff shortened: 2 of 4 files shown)


tests/conftest.py
```diff
@@ -57,3 +58,18 @@ def fake_fs():
...
+    if platform.system() == 'Windows':
+        monkeypatch.setenv('USERPROFILE', __tests__ + path)
+    else:
+        monkeypatch.setenv('HOME', __tests__ + path)
+
+
+def set_az_path(monkeypatch):
+    if platform.system() == 'Windows':
+        monkeypatch.setenv('Path', __tests__ + "\\testdata\\windows\\")
+        monkeypatch.setenv('COMSPEC', 'C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe')
```
mgyucht, line 73: I'm not familiar with this variable, what is it used for? Do we only need to set this environment variable when configuring Path?

tests/test_auth.py
```diff
@@ -2,7 +2,7 @@
 # In case of editing this file, make sure the change is propagated to all Databricks SDK codebases
 from databricks.sdk.core import Config
 
-from .conftest import __tests__, raises
+from .conftest import __tests__, raises, set_az_path, set_home
```
mgyucht, line 5: Is `__tests__` still needed?

tests/conftest.py
```diff
@@ -57,3 +58,18 @@ def fake_fs():
         patcher.fs.add_real_directory(test_data_path)
 
         yield patcher.fs # This will return a fake file system
+
+
+def set_home(monkeypatch, path):
```
mgyucht, line 63: Interesting that you don't need to change the direction of the slashes here?

mgyucht (commented):
To fix test_config_corrupt_config, we need to make two changes:
1. fix line 42 of conftest.py, where we expect forward slash as separator but should use `os.sep`.
2. The actual expected message in the at-raises annotation of the test needs to use backslashes instead of forward-slashes. I believe you can do something like:
```
msg = '...'
if os.name == 'nt':
  msg.replace('/', '\\')
@raises(msg)
def test_config_corrupt_config(...):
...
```

tests/conftest.py
parthban-db: This basically points to the command line interpreter by default windows program run on cmd. But we are running a powershell script to mock the `az` so we need to run it on powershell.

tests/test_auth.py
parthban-db: in line 219 of `test_auth.py`, We are pointing the `PATH` to some garbage place such that It will not find the `az` command. I am currently not passing any argument related to path in `set_az_path`. Because in rest of the cases the path is fixed. thats why I needed the `__tests__`. what are your suggestions regarding this?

tests/conftest.py
parthban-db: I didn't find anything on internet regarding this. But maybe it fixes the path before setting the variable.

parthban-db:
> To fix test_config_corrupt_config, we need to make two changes:
> 
> 1. fix line 42 of conftest.py, where we expect forward slash as separator but should use `os.sep`.
> 2. The actual expected message in the at-raises annotation of the test needs to use backslashes instead of forward-slashes. I believe you can do something like:
> 
> ```
> msg = '...'
> if os.name == 'nt':
>   msg.replace('/', '\\')
> @raises(msg)
> def test_config_corrupt_config(...):
> ...
> ```

Do I fix this issue in this PR? As the current change basically fix the Environment variable issue.

tests/conftest.py
mgyucht: Ah understood, and this is only needed for tests that run our mock `az` command.

tests/test_auth.py
mgyucht: It's alright, more of a nit not to import things that aren't used, but if it is still used I just missed it. We can leave as-is.

tests/conftest.py
```diff
@@ -39,7 +39,11 @@ def wrapper(*args, **kwargs):
             with pytest.raises(ValueError) as info:
                 func(*args, **kwargs)
             exception_str = str(info.value)
-            exception_str = exception_str.replace(__tests__ + '/', '')
+            if platform.system() == 'Windows':
+                exception_str = exception_str.replace(__tests__ + '\\', '')
+                exception_str = exception_str.replace('\\', '/')
```
mgyucht, line 44: This is a little bit weird since it means we will have to change all `\` to `/` in test assertions, which is why I preferred customizing the message itself depending on the OS. Not a big deal though, as these test cases change rarely and only one has a `\\` in the exception message now.

mgyucht (approved):
This is fine, thanks for the fixes!

PR #700, opened 2024-07-08
[Feature] Add DataPlane support

hectorcast-db:
## Changes
Add DataPlane support

## Tests
- [X] `make test` run locally
- [X] `make fmt` applied
- [X] Manual test against staging workspace (prod workspaces don't support DataPlane APIs)


.codegen/service.py.tmpl
```diff
@@ -338,6 +370,22 @@ self._api.do('{{.Verb}}',
   {{- end}}
   {{- if and .IsRequestByteStream .RequestBodyField }}, data={{template "safe-snake-name" .RequestBodyField}}{{ end }}
   {{- if .IsResponseByteStream }}, raw=True{{ end }})
+{{- else -}}
+  self._api.do('{{.Verb}}',
+    url=data_plane_details.endpoint_url
```
mgyucht, line 375: Small suggestion: I think we don't need to duplicate all of the parts of how we call the `do()` method here, since the only differences are in the URL and auth we supply. So something like
```
{{ if .Service.IsDataPlane }}
, url=data_plane_details.endpoint_url
, auth=auth
{{ else }}
{{ template "path" . }}
{{ end }}
```
would reduce duplication in this template.

databricks/sdk/core.py
```diff
@@ -133,31 +133,36 @@ def get_oauth_token(self, auth_details: str) -> Token:
...
            files=None,
            data=None,
+           auth=None,
            response_headers: List[str] = None) -> Union[dict, BinaryIO]:
-        # Remove extra `/` from path for Files API
-        # Once we've fixed the OpenAPI spec, we can remove this
-        path = re.sub('^/api/2.0/fs/files//', '/api/2.0/fs/files/', path)
         if headers is None:
             headers = {}
+        if url is None:
```
mgyucht, line 148: We do need to just remove this, but I think better not to do it in this PR.

databricks/sdk/core.py
```diff
@@ -133,31 +133,36 @@ def get_oauth_token(self, auth_details: str) -> Token:
...
-           path: str,
+           path: str = None,
+           url: str = None,
            query: dict = None,
            headers: dict = None,
            body: dict = None,
            raw: bool = False,
            files=None,
            data=None,
+           auth=None,
```
mgyucht, line 144: Can you add a type annotation here?

databricks/sdk/data_plane.py
```diff
@@ -0,0 +1,49 @@
...
+
+class DataPlaneService:
+
+    def __init__(self):
+        self._data_plane_info = {}
+        self._tokens = {}
+        self._lock = threading.Lock()
+
+    def get_data_plane_details(self, method: str, params: List[str], info_getter: Callable[[], DataPlaneInfo],
+                               refresh: Callable[[str], Token]):
```
mgyucht, line 23: Would you mind adding a doccomment?

databricks/sdk/data_plane.py
```diff
@@ -0,0 +1,49 @@
...
+from dataclasses import dataclass
+from typing import Callable, List
+
+from databricks.sdk.oauth import Token
+from databricks.sdk.service.oauth2 import DataPlaneInfo
+
+
+@dataclass
+class DataPlaneDetails:
+    endpoint_url: str
```
mgyucht, line 11: nit: could you add a doccomment?

.codegen/service.py.tmpl
```diff
@@ -195,7 +206,27 @@ class {{.PascalName}}API:{{if .Description}}
...
+            response = self._control_plane.{{.Service.DataPlaneInfoMethod.SnakeName}}(
+                {{- range .Service.DataPlaneInfoMethod.Request.Fields }}
+                {{.SnakeName}} = {{.SnakeName}},
+                {{- end}}
+                )
+            if response.{{(index .DataPlaneInfoFields 0).SnakeName}} is None:
+            	raise Exception("Resource does not support direct Data Plane access")
+            return response{{range .DataPlaneInfoFields}}.{{.SnakeName}}{{end}}
+
+        get_params = [{{- range .Service.DataPlaneInfoMethod.Request.Fields }}'{{.SnakeName}}',{{end}}]
```
mgyucht, line 222: I think this is a typo? Should be 
```suggestion
        get_params = [{{- range .Service.DataPlaneInfoMethod.Request.Fields }}{{.SnakeName}},{{end}}]
```

mgyucht (commented):
A few small nits, otherwise basically looks good to me!

mgyucht (approved):

PR #702, opened 2024-07-08
Fix test_core for windows

parthban-db:
## Changes

Fix `test_core.py` for windows. 

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -23,6 +23,7 @@ from databricks.sdk.credentials_provider import (CliTokenSource,
                                                  databricks_cli)
 from databricks.sdk.environments import (ENVIRONMENTS, AzureEnvironment, Cloud,
                                          DatabricksEnvironment)
+from databricks.sdk.oauth import Token
 from databricks.sdk.service.catalog import PermissionsChange
 from databricks.sdk.service.iam import AccessControlRequest
 from databricks.sdk.version import __version__
@@ -67,9 +68,14 @@ def test_databricks_cli_token_parse_expiry(date_string, expected):
 
 
 def write_small_dummy_executable(path: pathlib.Path):
-    cli = path.joinpath('databricks')
-    cli.write_text('#!/bin/sh\necho "hello world"\n')
-    cli.chmod(0o755)
+    if platform.system() == "Windows":
+        cli = path.joinpath('databricks.exe')
+        cli.touch()
+        cli.write_text('@echo off\necho "hello world"\n')
+    else:
+        cli = path.joinpath('databricks')
+        cli.write_text('#!/bin/sh\necho "hello world"\n')
+        cli.chmod(0o755)
     assert cli.stat().st_size < 1024
     return cli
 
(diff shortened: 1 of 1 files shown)


tests/test_core.py
```diff
@@ -175,10 +187,23 @@ def test_databricks_cli_credential_provider_installed_legacy(config, monkeypatch
...
+def test_databricks_cli_credential_provider_installed_new(config, monkeypatch, tmp_path, mocker):
+    get_mock = mocker.patch('databricks.sdk.credentials_provider.CliTokenSource.refresh',
+                            return_value=Token(access_token='token',
+                                               token_type='Bearer',
+                                               expiry=datetime(2023, 5, 22, 0, 0, 0)))
     write_large_dummy_executable(tmp_path)
-    monkeypatch.setenv('PATH', str(os.pathsep).join([tmp_path.as_posix(), os.environ['PATH']]))
+    path = str(os.pathsep).join([tmp_path.as_posix(), os.environ['PATH']])
+    if platform.system() == 'Windows':
+        path = pathlib.PureWindowsPath(path)
```
mgyucht, line 198: Are this and line 201 needed? I wonder because `pathlib.Path` should behave in a system-dependent way, i.e. when run on windows they will use windows path semantics and on mac/unix they will use mac/unix path semantics.

tests/test_core.py
```diff
@@ -175,10 +187,23 @@ def test_databricks_cli_credential_provider_installed_legacy(config, monkeypatch
...
+        monkeypatch.setenv('COMSPEC', 'C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe')
+    else:
+        path = pathlib.PurePosixPath(path)
+    path = path.__str__()
```
mgyucht, line 202: Note: just use `str(path)`, it's the same but more idiomatic for python.

tests/test_core.py
```diff
@@ -133,9 +139,15 @@ def test_databricks_cli_token_source_installed_legacy_with_symlink(config, monke
...
     dir2.mkdir()
 
-    (dir1 / "databricks").symlink_to(write_small_dummy_executable(dir2))
+    if platform.system() == 'Windows':
+        (dir1 / "databricks.exe").symlink_to(write_small_dummy_executable(dir2))
+        path = pathlib.PureWindowsPath(dir1)
+    else:
+        (dir1 / "databricks").symlink_to(write_small_dummy_executable(dir2))
+        path = pathlib.PurePosixPath(dir1)
+    path = path.__str__()
```
mgyucht, line 148: Ditto here, use `str(path)`.

mgyucht (approved):
Couple questions/small suggestions, once you've addressed them, feel free to merge.

tests/test_core.py
parthban-db: Thank you for pointing this out. I've addressed the changes accordingly.

PR #719, opened 2024-08-06
[Fix] Fix test_get_workspace_client and test_runtime_auth_from_jobs

mgyucht:
## Changes
This PR fixes the current failing integration tests for the Python SDK, unblocking their release.

There are two issues:
1. get_workspace_client fails in our integration tests because we call it with a workspace that is not UC-enabled. Because tests are authenticated as service principals, and it isn't possible to add account-level service principals to non-UC workspaces, this call fails. I address this by running this test against a UC-enabled workspace.
2. test_runtime_auth_from_jobs fails because a new LTS DBR version was released (15.4) that doesn't support DBFS library installations. To address this, I have created two tests: test_runtime_auth_from_jobs_dbfs, which tests native auth using the SDK installed from DBFS up to LTS 14.3, and test_runtime_auth_from_jobs_volumes, which does the same with the SDK installed from a volume. 

## Tests
All integration tests passed (retriggered the GCP integration test locally after adding single user data security mode).

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -60,6 +60,18 @@ def a(env_or_skip) -> AccountClient:
 
 
 @pytest.fixture(scope='session')
+def ucacct(env_or_skip) -> AccountClient:
+    _load_debug_env_if_runs_from_ide('ucacct')
+    env_or_skip("CLOUD_ENV")
+    account_client = AccountClient()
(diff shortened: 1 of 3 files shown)


renaudhartert-db (approved):
LGTM

PR #723, opened 2024-08-08
[Fix] Fix `DatabricksConfig.copy` when authenticated with OAuth

parthban-db:
## Changes

`DatabricksCliTokenSource().token()` itself can't be copied. So, Deep Copy can't be performed for Config.
Added the wrapper function which can be copied. So, Deep copy can be performed.
## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


nfx (approved):
confirmed it works locally. hardened integration test via https://github.com/databricks/databricks-sdk-py/pull/736

mgyucht (approved):

PR #725, opened 2024-08-13
[Internal] Update Jobs GetRun API to support paginated responses for jobs and ForEach tasks with >100 runs

gkiko10:
## Changes
Jobs GetRun 2.2 behind the scenes pagination

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied


databricks/sdk/mixins/jobs.py
```diff
@@ -0,0 +1,33 @@
...
+                              page_token=page_token)
+
+        isPaginatingIterations = run.iterations is not None and len(run.iterations) > 0
+
+        while run.next_page_token is not None:
+            next_run = super().get_run(run_id,
+                                       include_history=include_history,
+                                       include_resolved_values=include_resolved_values,
+                                       page_token=run.next_page_token)
+            if (isPaginatingIterations):
```
mgyucht, line 26:
```suggestion
            if isPaginatingIterations:
```

databricks/sdk/mixins/jobs.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class JobsExt(jobs.JobsAPI):
+
+    def get_run(self,
+                run_id: int,
+                *,
+                include_history: Optional[bool] = None,
+                include_resolved_values: Optional[bool] = None,
+                page_token: Optional[str] = None) -> jobs.Run:
```
mgyucht, line 13: Should we have a doccomment here explaining that this makes multiple requests to return all tasks/iterations when the underlying has more than 100?

tests/test_jobs_mixin.py
```diff
@@ -0,0 +1,112 @@
...
+        "next_page_token": "tokenToThirdPage",
+        "prev_page_token": "initialToken"
+    }
+    run3 = {"tasks": [{"run_id": 4}], "next_page_token": None, "prev_page_token": "tokenToSecondPage"}
+    requests_mock.get(make_path_pattern(1337, "initialToken"), text=json.dumps(run1))
+    requests_mock.get(make_path_pattern(1337, "tokenToSecondPage"), text=json.dumps(run2))
+    requests_mock.get(make_path_pattern(1337, "tokenToThirdPage"), text=json.dumps(run3))
+    w = WorkspaceClient(config=config)
+
+    run = w.jobs.get_run(1337, page_token="initialToken")
```
mgyucht, line 38: Will users ever specify an initial token?

tests/test_jobs_mixin.py
```diff
@@ -0,0 +1,112 @@
...
+        }],
+        "next_page_token": None,
+        "prev_page_token": "tokenToSecondPage"
+    }
+    requests_mock.get(make_path_pattern(1337, "initialToken"), text=json.dumps(run1))
+    requests_mock.get(make_path_pattern(1337, "tokenToSecondPage"), text=json.dumps(run2))
+    requests_mock.get(make_path_pattern(1337, "tokenToThirdPage"), text=json.dumps(run3))
+    w = WorkspaceClient(config=config)
+
+    run = w.jobs.get_run(1337, page_token="initialToken")
```
mgyucht, line 95: ditto about the initial token?

mgyucht (approved):
Overall LGTM, just a few small nits about style & test cases. The main question I have is whether we should add an extra documentation comment for this method.

tests/test_jobs_mixin.py
mgyucht: Discussed offline. Let's leave the parameter there so the mixin matches the underlying API. We'll clean up token parameters separately and can remove it here when we do that.

ricardoc-db (approved):

mgyucht (approved):

PR #729, opened 2024-08-16
[Fix] Fixed regression introduced in v0.30.0 causing `ValueError: Invalid semantic version: 0.33.1+420240816190912`

nfx:
## Changes

This PR fixes SemVer regex to follow the official recommendation to capture more patterns. It also ensures that patterns are both SemVer and PEP440 compliant.

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied


hectorcast-db (approved):

renaudhartert-db:
Could we add a link to the source of the regression?

PR #740, opened 2024-08-29
Accept both `str` and `Path` types for functions in `databricks/sdk/service/files.py`

fpgmaas:
## Changes

Currently, the functions in `databricks/sdk/service/files.py` accept only `str` for arguments that are paths. This can lead to confusing behavior for end-users when passing in a `Path` object.

## Tests
I parametrized the unit tests.

- [x] `make test` run locally
- [x] `make fmt` applied

### DIFF
--- a/databricks/sdk/service/files.py
+++ b/databricks/sdk/service/files.py
@@ -6,6 +6,8 @@ import logging
 from dataclasses import dataclass
 from typing import BinaryIO, Dict, Iterator, List, Optional
 
+from pathlib import Path
+
 from ._internal import _escape_multi_segment_path_parameter, _repeated_dict
 
 _LOG = logging.getLogger('databricks.sdk')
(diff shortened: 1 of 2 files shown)


mgyucht (changes requested):
Hi @fpgmaas, thanks for contributing this change to the SDK! Unfortunately, the code you're modifying is autogenerated and would be overwritten in the next SDK release if merged. Implementing this would require a change in our codegen tool itself. Can I ask you to make a feature request instead?

So that I understand the rationale: "This can lead to confusing behavior for end-users when passing in a Path object." The interfaces for these methods all specify that they accept strings. Can users just call `str(...)` themselves before using these methods?

fpgmaas:
> Hi @fpgmaas, thanks for contributing this change to the SDK! Unfortunately, the code you're modifying is autogenerated and would be overwritten in the next SDK release if merged. Implementing this would require a change in our codegen tool itself. Can I ask you to make a feature request instead?


Thanks for your response. I overlooked the comment `# Code generated from OpenAPI specs by Databricks SDK Generator. DO NOT EDIT.` at the top, my bad! I opened the issue https://github.com/databricks/databricks-sdk-py/issues/762.

> Can users just call str(...) themselves before using these methods?

Yes, that will definitely work. However, the error that is thrown is not very clear which makes it a bit annoying to debug. So for a better user experience I think allowing `Path` objects to the `path` argument would be a nice improvement.

renaudhartert-db:

PR #750, opened 2024-09-05
[Feature] Support Models in `dbutils.fs` operations

shichengzhou-db:
## Changes

- Support files operations in WorkspaceClient.Files for Databricks UC Model artifacts so that user can use databricks sdk to download UC model artifacts.
- This PR is part of the work to migrate mlflow client towards using databricks sdk for model artifacts download/upload operations for better security.

## Tests

- Existing tests in test_dbfs_mixins.py, similar to how _VolumesPath is tested
- The following code works
```
w = WorkspaceClient()
resp = w.files.download("/Models/system/ai/dbrx_instruct/3/MLmodel")
```

- [x] `make test` run locally
- [x] `make fmt` applied

### DIFF
--- a/databricks/sdk/mixins/files.py
+++ b/databricks/sdk/mixins/files.py
@@ -167,7 +167,7 @@ class _DbfsIO(BinaryIO):
         return f"<_DbfsIO {self._path} {'read' if self.readable() else 'write'}=True>"
 
 
-class _VolumesIO(BinaryIO):
+class _FilesIO(BinaryIO):
 
     def __init__(self, api: files.FilesAPI, path: str, *, read: bool, write: bool, overwrite: bool):
         self._buffer = []
@@ -262,7 +262,7 @@ class _VolumesIO(BinaryIO):
         self.close()
 
     def __repr__(self) -> str:
-        return f"<_VolumesIO {self._path} {'read' if self.readable() else 'write'}=True>"
+        return f"<_FilesIO {self._path} {'read' if self.readable() else 'write'}=True>"
 
 
 class _Path(ABC):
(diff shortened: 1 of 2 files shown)


databricks/sdk/mixins/files.py
```diff
@@ -467,6 +565,75 @@ def __repr__(self) -> str:
         return f'<_VolumesPath {self._path}>'
 
 
+class _ModelsPath(_Path):
```
mgyucht, line 568: Do we need a new Path implementation? It seems like this just uses the Files API under the hood. From what I can tell, the only difference is `open()`, which returns ModelsIO now. That is in turn almost identical to _FilesIO.

If we changed https://github.com/databricks/databricks-sdk-py/pull/750/files#diff-469b96ba8e49ef7cbe8d3b4577d2a0ddf3a3552638b76d87bfccb936fda65d18R759 to check whether the path started with either `/Volumes` or `/Models`, I think the existing implementation would work. Is this correct? If so, we should just rename that to `_FilesPath` (as it uses the Files API) and reuse it.

Ultimately, that `_path` method should probably use some kind of prefix allowlist to determine whether the `_FilesPath` should be used, and we can add `/Volumes` and `/Models` as the initial prefixes.

mgyucht (changes requested):
I think these basically duplicate _VolumesPath and _VolumesIO. Let's reuse those instead of defining duplicate implementations. We can rename those to `_FilesPath` and `_FilesIO` to indicate that they can be used generally for resources backed by the Files API.

databricks/sdk/mixins/files.py
shichengzhou-db: @mgyucht that sounds better. thanks for the suggestion, updated.

(I am not in the osp team but offered to do this work so we can speed up mlflow adoption of the databricks sdk, but I think since it's all Files API / Presigned URLs behind this sdk, we should be able to reuse the same path and io class for all securables)

mgyucht (approved):
LGTM, thanks for this contribution!

shichengzhou-db:
jenkins merge

shichengzhou-db:
hey @mgyucht how you know how to merge this pr? not seeing anything on my UI for merging

PR #761, opened 2024-09-13
[Feature] Integrate Databricks SDK with Model Serving Auth Provider

aravind-segu:
## Changes
This PR introduces a new model serving auth method to Databricks SDK. 
- If the correct environment variables are set to identify a model serving environment
- Check to see if there is an oauth file written by the serving environment
- If this file exists use the token here for authentication

## Tests
Added Unit tests

- [x] `make test` run locally
- [x] `make fmt` applied


databricks/sdk/credentials_provider.py
```diff
@@ -664,6 +665,86 @@ def inner() -> Dict[str, str]:
...
+    _MODEL_DEPENDENCY_OAUTH_TOKEN_FILE_PATH = "/var/credentials-secret/model-dependencies-oauth-token"
+
+    @classmethod
+    def is_in_databricks_model_serving_environment(cls) -> bool:
+        """
+        Check if the code is running in Databricks Model Serving environment.
+        The environment variable set by Databricks when starting the serving container.
+        """
+        val = (
+            os.environ.get("IS_IN_DB_MODEL_SERVING_ENV")
```
smurching, line 678: Could we test both of these env vars separately (maybe `pytest.mark.parametrize` the appropriate test to achieve this)? Having either should work right?

tests/test_model_serving_auth.py
```diff
@@ -0,0 +1,47 @@
...
+def test_model_serving_auth(monkeypatch):
+    ## In mlflow we check for these two environment variables to return the correct config
+    monkeypatch.setenv('IS_IN_DB_MODEL_SERVING_ENV', 'true')
+    monkeypatch.setenv('DB_MODEL_SERVING_HOST_URL', 'x')
+    # patch mlflow to read the file from the test directory
+    monkeypatch.setattr(
+        "databricks.sdk.credentials_provider.ModelServingAuthProvider._MODEL_DEPENDENCY_OAUTH_TOKEN_FILE_PATH",
+        "tests/testdata/model-serving-test-token")
+
+    cfg = Config()
```
smurching, line 13: +1 thanks for testing the interface & not implementation here

databricks/sdk/credentials_provider.py
```diff
@@ -664,6 +665,86 @@ def inner() -> Dict[str, str]:
...
+            or os.environ.get("IS_IN_DATABRICKS_MODEL_SERVING_ENV") or "false")
+        return val.lower() == "true"
+
+    @classmethod
+    def should_fetch_model_serving_environment_oauth(cls) -> bool:
+        """
+        Check whether this is the model serving environment
+        Additionally check if the oauth token file path exists
+        """
+        return (cls.is_in_databricks_model_serving_environment()
```
smurching, line 690: Could we test the different branches here? E.g. env var missing but file is present, and vice versa

databricks/sdk/credentials_provider.py
```diff
@@ -664,6 +665,86 @@ def inner() -> Dict[str, str]:
...
+
+    @classmethod
+    def get_databricks_host_token(cls) -> Optional[Tuple[str, str]]:
+        if not cls.should_fetch_model_serving_environment_oauth():
+            return None
+        MODEL_SERVING_HOST_ENV_VAR = "DATABRICKS_MODEL_SERVING_HOST_URL"
+        DB_MODEL_SERVING_HOST_ENV_VAR = "DB_MODEL_SERVING_HOST_URL"
+
+        # read from DB_MODEL_SERVING_HOST_ENV_VAR if available otherwise MODEL_SERVING_HOST_ENV_VAR
+        host = os.environ.get(DB_MODEL_SERVING_HOST_ENV_VAR) or os.environ.get(MODEL_SERVING_HOST_ENV_VAR)
```
smurching, line 716: Test both branches here?

databricks/sdk/credentials_provider.py
```diff
@@ -664,6 +665,86 @@ def inner() -> Dict[str, str]:
...
+@credentials_strategy('model-serving', [])
+def model_serving_auth(cfg: 'Config') -> Optional[CredentialsProvider]:
+    try:
+        host, token = ModelServingAuthProvider.get_databricks_host_token()
+        if token is None:
+            raise ValueError("Unable to Authenticate using Model Serving Environment")
+        if cfg.host is None:
+            cfg.host = host
+    except Exception as e:
+        logger.info("Unable to use Databricks Model Serving Environment " + str(e))
```
smurching, line 731: This is an unexpected case (we thought we should be able to authenticate with model serving auth but couldn't), right?

smurching: Should this be a `logger.warn` with `exc_info` passed (see https://docs.python.org/3/library/logging.html#logging.Logger.debug for details)?

smurching (commented):
A few comments, thanks @aravind-segu !

databricks/sdk/credentials_provider.py
```diff
@@ -664,6 +665,86 @@ def inner() -> Dict[str, str]:
...
+        return None
+
+    logger.info("Using Databricks Model Serving Authentication")
+
+    def inner() -> Dict[str, str]:
+        # Call here again to get the refreshed token
+        host, token = ModelServingAuthProvider.get_databricks_host_token()
```
mgyucht, line 738: This method is called with every request. Do we need to check the file every time?  We incur the cost of opening a file & reading its contents on every request. Is there an expiry included in the token file? We could use that to determine whether we need to refresh the token.

I tested a simple benchmark:

```py
import json
import timeit
import os

with open('sample.json', 'w') as f:
    json.dump([{'key': 'value'} for _ in range(100)], f)

def benchmark():
    with open('sample.json', 'r') as f:
        data = json.load(f)

print(f"Average time to open, read, and parse file: {timeit.timeit(benchmark, number=1000):.6f} seconds")
```
This came out to 0.07 seconds on my laptop. It might be nice to avoid this overhead, especially for latency-sensitive use-cases.

databricks/sdk/credentials_provider.py
```diff
@@ -664,6 +665,86 @@ def inner() -> Dict[str, str]:
...
+        # If the host is not declared, then get the host from implicit credential info in the
+        # model serving environment
+        if cfg.host is None:
+            cfg.host = host
```
mgyucht, line 742: This is already done on line 728-729. Do we need to repeat it here?

mgyucht (changes requested):
Mostly looking good, but we shouldn't throw exceptions from the credential strategy, otherwise DefaultCredentials will stop trying later auth mechanisms. This will be problematic if new strategies are added later. Otherwise, a few nits.

Were you able to test this out with a manual test?

databricks/sdk/credentials_provider.py
```diff
@@ -664,6 +665,86 @@ def inner() -> Dict[str, str]:
...
+
+        return (host, token)
+
+
+@credentials_strategy('model-serving', [])
+def model_serving_auth(cfg: 'Config') -> Optional[CredentialsProvider]:
+    try:
+        host, token = ModelServingAuthProvider.get_databricks_host_token()
+        if token is None:
+            raise ValueError("Unable to Authenticate using Model Serving Environment")
```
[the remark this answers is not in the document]
aravind-segu: I am raising an error here, but then catching it immediately and returning None after. I wanted to add an extra message to differentiate whether reading the token failed or whether the token read was itself none.

databricks/sdk/credentials_provider.py
```diff
@@ -664,6 +665,86 @@ def inner() -> Dict[str, str]:
     return inner
 
 
+class ModelServingAuthProvider():
+    _MODEL_DEPENDENCY_OAUTH_TOKEN_FILE_PATH = "/var/credentials-secret/model-dependencies-oauth-token"
```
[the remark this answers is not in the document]
aravind-segu: Not really. We have it in both the mlflow code and model serving code to write and read from this file

databricks/sdk/credentials_provider.py
aravind-segu: Good call, added a 5 min cache

aravind-segu:
> Mostly looking good, but we shouldn't throw exceptions from the credential strategy, otherwise DefaultCredentials will stop trying later auth mechanisms. This will be problematic if new strategies are added later. Otherwise, a few nits.
> 
> Were you able to test this out with a manual test?

Manual Test Endpoint Here: https://e2-dogfood.staging.cloud.databricks.com/ml/endpoints/agents_main-default-test_agent_09_16?o=6051921418418893

aravind-segu:
@mgyucht We discussed in the call about making this a `oauth-credential-provider`. However that requires a `refresht_token` we do not have. We can set that to None but what happens then when the `expiry` is hit. 

In terms of direct ingress, I do not think we are going to be hitting Model Serving Direct Ingress Endpoints right now. The only thing that we could do is [talk to a Foundational Model Serving Endpoint](https://github.com/mlflow/mlflow/blob/e5d1280a467f0b51692ff3edf29e191fc52f36ef/mlflow/deployments/databricks/__init__.py#L122) through the workspace client. Currently we already do that in our Mlflow code by sending a request [directly](https://github.com/mlflow/mlflow/blob/e5d1280a467f0b51692ff3edf29e191fc52f36ef/mlflow/utils/rest_utils.py#L187C1-L202C17). 

WIth this change we will now start to use databricks sdk workspace client to make this call. We will use the code [here](https://github.com/mlflow/mlflow/blob/e5d1280a467f0b51692ff3edf29e191fc52f36ef/mlflow/utils/rest_utils.py#L110). As it already worked before, I am assuming that it will work again as this directly calls the api client. Do we still need to integrate `oauth-credential-provider` now.

databricks/sdk/credentials_provider.py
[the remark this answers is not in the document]
smurching: Yeah I guess to be clear, there is already public OSS code in MLflow that references this path, but we don't document it anywhere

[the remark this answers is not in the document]
smurching: Chatted offline, @aravind-segu  will leave a code comment referencing back the source MLflow code that we're copying + slightly-modifying here

databricks/sdk/credentials_provider.py
```diff
@@ -664,6 +665,80 @@ def inner() -> Dict[str, str]:
...
+        return (host, token)
+
+
+@credentials_strategy('model-serving', [])
+def model_serving_auth(cfg: 'Config') -> Optional[CredentialsProvider]:
+    try:
+        model_serving_auth_provider = ModelServingAuthProvider()
+        host, token = model_serving_auth_provider.get_databricks_host_token()
+        if token is None:
+            raise ValueError("Unable to Authenticate using Model Serving Environment since the Token is None")
```
smurching, line 725: We shouldn't raise here right?

smurching: Oh I see we catch it further down

smurching, line 725:
```suggestion
            raise ValueError("Got malformed auth (empty token) when fetching auth implicitly available in Model Serving Environment. Please contact Databricks support")
```

Updating the message for clarity. Technically auth could still succeed from some later provider in the chain, but this is an unexpected case that will probably/usually result in an error, and which we should get notified about

databricks/sdk/credentials_provider.py
```diff
@@ -664,6 +665,80 @@ def inner() -> Dict[str, str]:
...
+        if cfg.host is None:
+            cfg.host = host
+    except Exception as e:
+        logger.warning("Unable to use Databricks Model Serving Environment", exc_info=e)
```
smurching, line 729:
```suggestion
        logger.warning("Unable to get auth from Databricks Model Serving Environment", exc_info=e)
```

smurching (commented):
Last comments, then LGTM!

smurching (approved):
LGTM!

databricks/sdk/credentials_provider.py
```diff
@@ -664,6 +665,84 @@ def inner() -> Dict[str, str]:
...
+
+        is_in_model_serving_env = (os.environ.get("IS_IN_DB_MODEL_SERVING_ENV")
+                                   or os.environ.get("IS_IN_DATABRICKS_MODEL_SERVING_ENV") or "false")
+        return (is_in_model_serving_env == "true"
+                and os.path.isfile(self.__class__._MODEL_DEPENDENCY_OAUTH_TOKEN_FILE_PATH))
+
+    def get_model_dependency_oauth_token(self, should_retry=True) -> str:
+        if self.current_token is not None and self.expiry_time > time.time():
+            return self.current_token
+        else:
```
mgyucht, line 692: nit: no need for an else branch if you return from the if branch.

databricks/sdk/credentials_provider.py
```diff
@@ -664,6 +665,84 @@ def inner() -> Dict[str, str]:
...
+            "DB_MODEL_SERVING_HOST_URL")
+        token = self.get_model_dependency_oauth_token()
+
+        return (host, token)
+
+
+@credentials_strategy('model-serving', [])
+def model_serving_auth(cfg: 'Config') -> Optional[CredentialsProvider]:
+    try:
+        model_serving_auth_provider = ModelServingAuthProvider()
```
mgyucht, line 724: Suggestion:
```suggestion
        model_serving_auth_provider = ModelServingAuthProvider()
        if not model_serving_auth_provider.should_fetch_model_serving_environment_oauth():
            logger.debug("model-serving: not in Databricks Model Serving, skipping")
            return None
```
Otherwise, we log a scary error message when this auth type is used outside of model serving environment.

databricks/sdk/credentials_provider.py
```diff
@@ -664,6 +665,84 @@ def inner() -> Dict[str, str]:
...
+    def should_fetch_model_serving_environment_oauth(self) -> bool:
+        """
+        Check whether this is the model serving environment
+        Additionally check if the oauth token file path exists
+        """
+
+        is_in_model_serving_env = (os.environ.get("IS_IN_DB_MODEL_SERVING_ENV")
+                                   or os.environ.get("IS_IN_DATABRICKS_MODEL_SERVING_ENV") or "false")
+        return (is_in_model_serving_env == "true"
+                and os.path.isfile(self.__class__._MODEL_DEPENDENCY_OAUTH_TOKEN_FILE_PATH))
```
mgyucht, line 687: You don't need `__class__` to refer to class variables in instance methods.
```suggestion
                and os.path.isfile(self._MODEL_DEPENDENCY_OAUTH_TOKEN_FILE_PATH))
```

databricks/sdk/credentials_provider.py
```diff
@@ -664,6 +665,84 @@ def inner() -> Dict[str, str]:
...
+
+    def get_model_dependency_oauth_token(self, should_retry=True) -> str:
+        if self.current_token is not None and self.expiry_time > time.time():
+            return self.current_token
+        else:
+            try:
+                with open(self.__class__._MODEL_DEPENDENCY_OAUTH_TOKEN_FILE_PATH) as f:
```
mgyucht, line 694: Ditto
```suggestion
                with open(self._MODEL_DEPENDENCY_OAUTH_TOKEN_FILE_PATH) as f:
```

databricks/sdk/credentials_provider.py
```diff
@@ -664,6 +665,84 @@ def inner() -> Dict[str, str]:
...
+                    oauth_dict = json.load(f)
+                    self.current_token = oauth_dict["OAUTH_TOKEN"][0]["oauthTokenValue"]
+                    self.expiry_time = time.time() + self.refresh_duration
+            except Exception as e:
+                # sleep and retry in case of any race conditions with OAuth refreshing
+                if should_retry:
+                    time.sleep(0.5)
```
mgyucht, line 701: Can you log the exception here? This will be useful debugging information if something goes wrong.

mgyucht (commented):
Close, just a few small comments.

databricks/sdk/credentials_provider.py
aravind-segu: Good call, we dont want that error message if it is not in the mdoel serving environment.

mgyucht (approved):

aravind-segu:
jenkins merge

PR #766, opened 2024-09-17
[Internal] Update SDK to latest OpenAPI spec

tanmay-db:
## Changes

Updating SDK to latest OpenAPI spec + fix generation (need to import `Optional`)

Note: `test_github_oidc_flow_works_with_azure` fails on genkit generate-sdk py but passes when run separately right after generation without any change. This seems to be non blocker so going ahead with SDK generation.  

## Tests

Unit tests. Nightly tests will run over release PR.
- [ ] `make test` run locally
- [ ] `make fmt` applied


.gitattributes
```diff
@@ -20,278 +20,3 @@ databricks/sdk/service/sharing.py linguist-generated=true
 databricks/sdk/service/sql.py linguist-generated=true
 databricks/sdk/service/vectorsearch.py linguist-generated=true
 databricks/sdk/service/workspace.py linguist-generated=true
-examples/account/billable_usage/download_usage_download.py linguist-generated=true
```
tanmay-db, line 23: This is the same issue we had for other SDKs for which we have created a backlog ticket to track.

renaudhartert-db (approved):

PR #767, opened 2024-09-18
Fix library error reference

kishaningithub:
## Changes

Fix error reference to point to this library instead of the equivalent go version

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied


renaudhartert-db:
Hi @kishaningithub, thank you for the PR! 

Before we can accept and merge it, we would need two things from you:

1. Our repository only accepts [signed commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification), could you back sign the commit so that these appear as "verified"?
2. We would need you to consent with our [DCO](https://github.com/databricks/databricks-sdk-py/blob/main/DCO). You can easily do that by following the steps in our [contribution guidelines](https://github.com/databricks/databricks-sdk-py/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

I'll proceed with approving and merging the PR once these two things are done.

Thanks!

kishaningithub:
I think https://github.com/databricks/databricks-sdk-py/pull/799 is a better fix for this than this current PR

PR #781, opened 2024-10-05
[Fix] Fix Model Serving Test

parthban-db:
## Changes

Fix Model Serving Tests. 
- Added preffered auth_type so that test will not use other auth_types first.
- Patch function that reads `.databrickscfg`.
- Unsetting already existing environment variables.

## Tests

- [x] `make test` run locally
- [x] `make fmt` applied


tanmay-db (approved):
synced offline, LGTM

PR #785, opened 2024-10-07
[Internal] Refactor ApiClient into `_BaseClient` and `ApiClient`

mgyucht:
## Changes
`ApiClient` is also coupled to the `Config` object, which means that it can't be used in situations where there is no config. For example, when fetching OIDC endpoints, the user may not have a complete `Config` instance yet. However, failures when requesting from those endpoints should still be retried according to the SDK's retry policy.

To address this, I've split the ApiClient into `_BaseClient` and `ApiClient`. `_BaseClient` is the core implementation of the client without any dependency on the `Config`. This is similar to what @rauchy did in the Java SDK to cut the dependency between the `ApiClient` and `DatabricksConfig`. The `_BaseClient` can then be used when fetching OIDC endpoint information.

This will be used in https://github.com/databricks/databricks-sdk-py/pull/784 to support retrying OAuth OIDC endpoint fetches.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied

### DIFF
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -18,7 +18,6 @@ JWT_BEARER_GRANT_TYPE = "urn:ietf:params:oauth:grant-type:jwt-bearer"
 OIDC_TOKEN_PATH = "/oidc/v1/token"
 
 
-
 class ApiClient:
 
     def __init__(self, cfg: Config):
--- a/databricks/sdk/core.py
+++ b/databricks/sdk/core.py
@@ -1,19 +1,13 @@
 import re
-from datetime import timedelta
-from types import TracebackType
-from typing import Any, BinaryIO, Iterator, Type
+from typing import BinaryIO
 from urllib.parse import urlencode
 
-from requests.adapters import HTTPAdapter
-
-from .casing import Casing
+from ._base_client import _BaseClient
 from .config import *
 # To preserve backwards compatibility (as these definitions were previously in this module)
 from .credentials_provider import *
-from .errors import DatabricksError, _ErrorCustomizer, _Parser
-from .logger import RoundTrip
+from .errors import DatabricksError, _ErrorCustomizer
 from .oauth import retrieve_token
-from .retries import retried
 
 __all__ = ['Config', 'DatabricksError']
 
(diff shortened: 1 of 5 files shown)


tests/fixture_server.py
```diff
@@ -0,0 +1,31 @@
+import contextlib
+import functools
+import typing
+from http.server import BaseHTTPRequestHandler
+
+
+@contextlib.contextmanager
+def http_fixture_server(handler: typing.Callable[[BaseHTTPRequestHandler], None]):
```
mgyucht, line 8: Moved here from `test_core.py`.

tests/test_base_client.py
```diff
@@ -0,0 +1,278 @@
...
+
+from databricks.sdk import errors, useragent
+from databricks.sdk._base_client import _BaseClient, _StreamingResponse
+from databricks.sdk.core import DatabricksError
+
+from .clock import FakeClock
+from .fixture_server import http_fixture_server
+
+
+class DummyResponse(requests.Response):
```
mgyucht, line 15: Moved here from `test_core.py`.

tests/test_base_client.py
```diff
@@ -0,0 +1,278 @@
...
+                           },
+                       }])),
+    (401, {}, {
+        'error_code': 'UNAUTHORIZED',
+        'message': 'errorMessage',
+    }, errors.Unauthenticated('errorMessage', error_code='UNAUTHORIZED')),
+    (403, {}, {
+        'error_code': 'FORBIDDEN',
+        'message': 'errorMessage',
+    }, errors.PermissionDenied('errorMessage', error_code='FORBIDDEN')),
```
mgyucht, line 95: These cases are slightly different because the ApiClient adds a debug string based on the Config when handling 401 and 403 errors. I haven't included that in this PR because that could induce a dependency on the Config.

tests/test_core.py
```diff
@@ -359,81 +297,25 @@ def test_deletes(config, requests_mock):
...
+         'error_code': 'FORBIDDEN',
+         'message': 'errorMessage',
+     },
+      errors.PermissionDenied('errorMessage. Config: host=http://localhost, auth_type=noop',
+                              error_code='FORBIDDEN')), ])
 def test_error(config, requests_mock, status_code, headers, body, expected_error):
     client = ApiClient(config)
     requests_mock.get("/test", json=body, status_code=status_code, headers=headers)
     with pytest.raises(DatabricksError) as raised:
-        client._perform("GET", "http://localhost/test", headers={"test": "test"})
```
mgyucht, line 436: Note that after this PR, it isn't possible to call `ApiClient._perform()` anymore.

databricks/sdk/core.py
```diff
@@ -25,53 +19,19 @@
...
+                                       header_factory=cfg.authenticate,
+                                       max_connection_pools=cfg.max_connection_pools,
+                                       max_connections_per_pool=cfg.max_connections_per_pool,
+                                       pool_block=True,
+                                       http_timeout_seconds=cfg.http_timeout_seconds,
+                                       extra_error_customizers=[_AddDebugErrorCustomizer(cfg)],
+                                       clock=cfg.clock)
 
     @property
     def account_id(self) -> str:
```
rauchy, line 37: What differentiates `account_id` and `is_account_client` from the rest of the things moved upwards? They're unused here so I'm assuming they are publicly facing, so I'm just curious on how to decide on which config attributes should be decoupled from `ApiClient` and which needs to stay.

rauchy (approved):
Looks great, just a small non-blocking question.

databricks/sdk/core.py
mgyucht: The way I thought about it: account_id and is_account_client are not inherently necessary for making API requests. In essence, they are really parameters of the config that happen to be in the ApiClient as well. The current architecture of the SDK doesn't expose the config directly to the *Api classes, so config-level attributes are "exposed" through the ApiClient class instead. However, fixing this would involve more invasive changes in the codegen process, which aren't necessary for now and will probably be revisited anyways with the next SDK version.

It doesn't seem like `is_account_client` is used within the SDK today, but it is publicly exposed. This is strongly coupled to the Config, because it requires knowledge of a host + logic to compute whether the provided host is an accounts host, which currently resides in the Config. In order to not break any callers, I left it as-is.

All of the other parameters are actually used inside of `_BaseClient` and affect its behavior, so they need to be included at that level.

PR #799, opened 2024-10-23
Minor fix in error message for generic error

dsculptor:
## Changes
Minor fix in error message for generic error

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied

Same as: https://github.com/databricks/databricks-sdk-py/pull/767


databricks/sdk/errors/parser.py
```diff
@@ -38,7 +38,7 @@ def _unknown_error(response: requests.Response) -> str:
     return (
         'This is likely a bug in the Databricks SDK for Python or the underlying '
         'API. Please report this issue with the following debugging information to the SDK issue tracker at '
-        f'https://github.com/databricks/databricks-sdk-go/issues. Request log:```{request_log}```')
```
mgyucht, line 41: Whoops, nice catch!

mgyucht (commented):
Thanks @dsculptor.

Let's extend this with one more change. In particular, I want to ensure that any error that is not parsed by any of the current error parsers is still handled in a graceful manner. To do this, let's implement a new error parser that is a catch-all error parser that just sets the entire response body as the message. That way, we never need to use this message in the error directly, as this error parser will always succeed. However, we can still log a message saying that we were unable to parse the error and fell back to this catch-all case.

renaudhartert-db:

PR #810, opened 2024-10-31
[Fix] Handle PySparkRuntimeError when populating globals in databricks.sdk.runtime

mgyucht:
## Changes
The runtime auth methods fail when using the Python SDK in a Python subprocess. In this environment, the `UserNamespaceInitializer` is present as it is built into DBR, but it depends on process-level state that is only initialized for the notebook process itself. This PR tolerates the error thrown by PySpark so that default auth can fall back to other auth methods in this environment.

## Tests

- [ ] `make test` run locally
- [ ] `make fmt` applied


renaudhartert-db:

PR #817, opened 2024-11-05
[Feature] Read streams by 1MB chunks by default.

renaudhartert-db:

This PR changes the `_BaseClient` to read streams by chunks of 1MB by default. 1MB was chosen as a good compromise between speed and memory usage (see PR #319). 

Note that this is not a new feature per se as it was possible to configure chunk size on the returned `_StreamResponse` before calling its read method. However, the functionality was not easy to discover and led several users to experience memory issues. The new default behavior is more defensive.

## How is this tested?

Added a few test cases to verify that streams are chunked as expected.


tests/test_base_client.py
```diff
@@ -276,3 +277,38 @@ def inner(h: BaseHTTPRequestHandler):
...
     assert len(requests) == 2
+
+
+@pytest.mark.parametrize('chunk_size,expected_chunks,data_size',
+                         [(5, 20, 100), # 100 / 5 bytes per chunk = 20 chunks
+                          (10, 10, 100), # 100 / 10 bytes per chunk = 10 chunks
+                          (200, 1, 100), # 100 / 200 bytes per chunk = 1 chunk
+                          ])
+def test_streaming_response_chunk_size(chunk_size, expected_chunks, data_size):
+    test_data = b"0" * data_size
```
ksafonov-db, line 288: Does it make sense to generate a variable test data (e.g. random with the fixed seed or a hard-coded block of text) so we have a higher confidence that bytes/chunks did not get mixed?

databricks/sdk/_base_client.py
```diff
@@ -68,6 +69,8 @@ def __init__(self,
         :param extra_error_customizers:
         :param debug_headers: Whether to include debug headers in the request log.
         :param clock: Clock object to use for time-related operations.
+        :param streaming_buffer_size: The size of the buffer to use for streaming responses. If None, all the
+            response content is loaded into memory at once.
```
ksafonov-db, line 73: Is this actually practical to allow reading all into memory? If that's achievable by passing big enough buffer size, do we need to make this parameter optional?

pietern:
Related PR: #319.

@renaudhartert-db Looking at the related PR, it's the responsibility of the caller to enable chunking or not. Was there a specific issue or regression you saw where this didn't happen or do you just want to do the right thing by default here?

renaudhartert-db:
> Related PR: #319.
> 
> @renaudhartert-db Looking at the related PR, it's the responsibility of the caller to enable chunking or not. Was there a specific issue or regression you saw where this didn't happen or do you just want to do the right thing by default here?

Thanks for calling this out, I've updated the PR description to clarify the intent.

tests/test_base_client.py
renaudhartert-db: Agreed, done.

databricks/sdk/_base_client.py
renaudhartert-db: My original intent was to provide a way to keep the current behavior. Though, I agree this is something we might want to simply remove for clarity. One can achieve a similar behavior by setting the size to an arbitrarily large number if needed.

databricks/sdk/_base_client.py
```diff
@@ -68,6 +69,7 @@ def __init__(self,
         :param extra_error_customizers:
         :param debug_headers: Whether to include debug headers in the request log.
         :param clock: Clock object to use for time-related operations.
+        :param streaming_buffer_size: The size of the buffer to use for streaming responses (default: 1MB).
```
pietern, line 72: Nit: isn't the default value included in the docs already? Less risk for it to go out of sync.

pietern (approved):

databricks/sdk/_base_client.py
renaudhartert-db: 👍

PR #821, opened 2024-11-13
[Fix] Rewind seekable streams before retrying

renaudhartert-db:
## How is this tested?

Added unit tests to verify that (i) non-seekable streams are not retried, and (ii) seekable streams are properly reset before retrying.

### DIFF
--- a/databricks/sdk/_base_client.py
+++ b/databricks/sdk/_base_client.py
@@ -1,3 +1,4 @@
+import io
 import logging
 import urllib.parse
 from datetime import timedelta
@@ -130,6 +131,14 @@ class _BaseClient:
         flattened = dict(flatten_dict(with_fixed_bools))
         return flattened
 
+    @staticmethod
+    def _is_seekable_stream(data) -> bool:
+        if data is None:
+            return False
+        if not isinstance(data, io.IOBase):
+            return False
+        return data.seekable()
+
     def do(self,
            method: str,
            url: str,
@@ -144,18 +153,25 @@ class _BaseClient:
         if headers is None:
             headers = {}
         headers['User-Agent'] = self._user_agent_base
-        retryable = retried(timeout=timedelta(seconds=self._retry_timeout_seconds),
-                            is_retryable=self._is_retryable,
-                            clock=self._clock)
-        response = retryable(self._perform)(method,
-                                            url,
-                                            query=query,
-                                            headers=headers,
-                                            body=body,
-                                            raw=raw,
-                                            files=files,
-                                            data=data,
-                                            auth=auth)
+
+        # Only retry if the request is not a stream or if the stream is seekable and
+        # we can rewind it. This is necessary to avoid bugs where the retry doesn't
+        # re-read already read data from the body.
+        call = self._perform
+        if data is None or self._is_seekable_stream(data):
+            call = retried(timeout=timedelta(seconds=self._retry_timeout_seconds),
+                           is_retryable=self._is_retryable,
+                           clock=self._clock)(call)
+
+        response = call(method,
+                        url,
+                        query=query,
+                        headers=headers,
+                        body=body,
+                        raw=raw,
+                        files=files,
+                        data=data,
+                        auth=auth)
 
         resp = dict()
         for header in response_headers if response_headers else []:
@@ -226,6 +242,12 @@ class _BaseClient:
                  files=None,
                  data=None,
                  auth: Callable[[requests.PreparedRequest], requests.PreparedRequest] = None):
+        # Keep track of the initial position of the stream so that we can rewind it if
+        # we need to retry the request.
+        initial_data_position = 0
+        if self._is_seekable_stream(data):
+            initial_data_position = data.tell()
+
         response = self._session.request(method,
                                          url,
                                          params=self._fix_query_string(query),
@@ -237,9 +259,15 @@ class _BaseClient:
                                          stream=raw,
                                          timeout=self._http_timeout_seconds)
         self._record_request_log(response, raw=raw or data is not None or files is not None)
+
         error = self._error_parser.get_api_error(response)
         if error is not None:
+            # If the request body is a seekable stream, rewind it so that it is ready
+            # to be read again in case of a retry.
+            if self._is_seekable_stream(data):
+                data.seek(initial_data_position)
             raise error from None
+
         return response
 
     def _record_request_log(self, response: requests.Response, raw: bool = False) -> None:
--- a/tests/test_base_client.py
+++ b/tests/test_base_client.py
@@ -1,3 +1,4 @@
+import io
 import random
 from http.server import BaseHTTPRequestHandler
 from typing import Iterator, List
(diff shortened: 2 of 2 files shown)


databricks/sdk/_base_client.py
```diff
@@ -226,6 +242,12 @@ def _perform(self,
                  files=None,
                  data=None,
                  auth: Callable[[requests.PreparedRequest], requests.PreparedRequest] = None):
+        # Keep track of the initial position of the stream so that we can rewind it if
+        # we need to retry the request.
```
pietern, line 246: Can you include an example of where the stream might not be at `0`?

pietern (approved):

databricks/sdk/_base_client.py
ksafonov-db: Does this need to be a shared variable between all retry attempts (initialized before the first attempt)?

databricks/sdk/_base_client.py
```diff
@@ -237,9 +259,15 @@ def _perform(self,
                                          stream=raw,
                                          timeout=self._http_timeout_seconds)
         self._record_request_log(response, raw=raw or data is not None or files is not None)
+
         error = self._error_parser.get_api_error(response)
         if error is not None:
+            # If the request body is a seekable stream, rewind it so that it is ready
+            # to be read again in case of a retry.
+            if self._is_seekable_stream(data):
+                data.seek(initial_data_position)
```
ksafonov-db, line 268: Nit: we're doing seek during the last retry attempt (after handling failure), in which case we'll not use this stream after seeking. If seek is a costly operation, this may matter.

tests/test_base_client.py
```diff
@@ -314,3 +315,98 @@ def mock_iter_content(chunk_size):
...
+        # Read some data from the stream first to verify that the stream is
+        # reset to the correct position rather than to its beginning.
+        stream.read(4)
+        assert stream.tell() == 4
+
+        # Should fail but reset the stream.
+        with pytest.raises(DatabricksError):
+            client._perform('POST', f'{host}/foo', data=stream)
+
+        assert received_data == [b"456789"]
```
ksafonov-db, line 375: If we made a few retry attempts, should we expect multiple entries in `received_data` array? (each being `b"456789"`)

tests/test_base_client.py
```diff
@@ -314,3 +315,98 @@ def mock_iter_content(chunk_size):
...
+        h.send_response(429)
+        h.send_header('Retry-After', '1')
+        h.end_headers()
+
+    stream = io.BytesIO(b"0123456789") # seekable stream
+
+    with http_fixture_server(inner) as host:
+        client = _BaseClient()
+
+        # Read some data from the stream first to verify that the stream is
```
ksafonov-db, line 366: nit: we could probably move this block out of `with http_fixture_server`

databricks/sdk/_base_client.py
renaudhartert-db: This could be needed if users use a single stream to produce a sequence of requests (e.g. upload a large file into chunks). Though, I don't know if we have such use cases at the moment. My real motivation was to be defensive here.

tests/test_base_client.py
```diff
@@ -314,3 +315,98 @@ def mock_iter_content(chunk_size):
...
+        # Should raise error immediately without retry.
+        with pytest.raises(DatabricksError):
+            client.do('POST', f'{host}/foo', data=stream)
+
+        # Verify that only one request was made (no retries).
+        assert len(requests) == 1
+        assert requests[0] == b"test data"
+
+
+def test_perform_resets_seekable_stream_on_error():
```
ksafonov-db, line 348: nit: maybe perform_resets_seekable_stream_on_retry?

databricks/sdk/_base_client.py
```diff
@@ -130,6 +131,14 @@ def flatten_dict(d: Dict[str, Any]) -> Dict[str, Any]:
         flattened = dict(flatten_dict(with_fixed_bools))
         return flattened
 
+    @staticmethod
+    def _is_seekable_stream(data) -> bool:
```
ksafonov-db, line 135: We might also want to add a test for this method, passing common types: string, byte array, buffered reader etc.

databricks/sdk/_base_client.py
renaudhartert-db: > Does this need to be a shared variable between all retry attempts (initialized before the first attempt)?

I don't think it does in this case because we systematically unwind at the end of the call. The benefit of this approach is that it keeps this change pretty minimal. The disadvantage is that we make on too many seek call (see other comment).

databricks/sdk/_base_client.py
renaudhartert-db: Agreed that this is a weakness of this solution. The alternative is to implement the unwind logic within the retry logic. Unfortunately, the current code (with the decorator) does not make this easy. 

My recommendation would be to proceed with this PR first to fix the bug, and clean-up in a follow-up PR focused on optimization. Please feel free to disagree.

tests/test_base_client.py
renaudhartert-db: Done.

databricks/sdk/_base_client.py
```diff
@@ -144,18 +153,25 @@ def do(self,
...
-                                            files=files,
-                                            data=data,
-                                            auth=auth)
+
+        # Only retry if the request is not a stream or if the stream is seekable and
+        # we can rewind it. This is necessary to avoid bugs where the retry doesn't
+        # re-read already read data from the body.
+        call = self._perform
+        if data is None or self._is_seekable_stream(data):
+            call = retried(timeout=timedelta(seconds=self._retry_timeout_seconds),
```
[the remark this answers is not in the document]
renaudhartert-db: Good point, done.

databricks/sdk/_base_client.py
renaudhartert-db: Done.

tests/test_base_client.py
renaudhartert-db: Correct, added a multiple calls so that this is tested.

databricks/sdk/_base_client.py
ksafonov-db: This just feels a bit awkward to me although I don't have a strong example. Could a customer rely on the resulting stream position in their code?

For the solution, we could pass a lambda to the retry function that would call it before each retry attempt. This lambda could even return e.g. false in case stream is not seekable, so that retry logic bubbles up the error.

tests/test_base_client.py
```diff
@@ -314,3 +315,140 @@ def mock_iter_content(chunk_size):
...
     assert len(content_chunks) == expected_chunks # correct number of chunks
     assert all(len(c) <= chunk_size for c in content_chunks) # chunks don't exceed size
+
+
+def test_is_seekable_stream():
+    client = _BaseClient()
+
+    # Test various input types that are not streams.
+    assert not client._is_seekable_stream(None) # None
+    assert not client._is_seekable_stream("string data") # str
```
ksafonov-db, line 325: Should we internally wrap static strings and byte arrays via `BytesIO` to make them seekable and so easily uploadable?

tests/test_base_client.py
```diff
@@ -314,3 +315,140 @@ def mock_iter_content(chunk_size):
...
+    # Read some data from the stream first to verify that the stream is
+    # reset to the correct position rather than to its beginning.
+    stream.read(4)
+    assert stream.tell() == 4
+
+    with http_fixture_server(inner) as host:
+        client = _BaseClient()
+
+        # Each call should fail and reset the stream.
+        with pytest.raises(DatabricksError):
```
ksafonov-db, line 410: In my previous comment about `received_data` I missed that we're actually calling the low-level `perform()` (which does not retry), not higher-level `do()`. I', thinking if it makes sense to test at the `do()` level, which would cover all the logic of retries and seeks?

ksafonov-db (commented):
I'd personally prefer if we were rewinding stream only when we need to. @pietern WDYT?

pietern:
@ksafonov-db I doubt it matters much. Regardless of the outcome of the operation, the caller will have to seek to a known location anyway if they want to keep the handle open in the first place.

Performance impact on failure is negligible because seek only updates the location offset in the file handle and doesn't trigger I/O directly (not counting readahead that the OS might do, but we can ignore that here).

I'll leave it to you to figure out if there is a case for pushing this down into a "pre-run" callback from a structural pov.

tests/test_base_client.py
renaudhartert-db: Done.

renaudhartert-db:
Confirmed with @ksafonov-db offline that he is good with the current state of the PR

Issue #844, opened 2024-12-30
[FEATURE] Recover on download failures


ksafonov-db:
**Problem Statement**
Currently download using Files API client fails on network error or server timeout.

**Proposed Solution**
Make Files API client automatically re-connect to the server and resume the download in case of temporary failures.

PR #845, opened 2024-12-30
Files API client: recover on download failures (#844)

ksafonov-db:

1. Extending Files API client to support resuming download on failures. New implementation tracks current offset in the input stream and issues a new download request from this point in case of an error.
2. New code path is enabled by 'DATABRICKS_ENABLE_EXPERIMENTAL_FILES_API_CLIENT' config parameter.

## How is this tested?

Added unit tests for the new code path:
`% python3 -m pytest tests/test_files.py`


renaudhartert-db (approved):
LGTM

PR #850, opened 2025-01-06
[Internal] Migrate workflows that need write access to use hosted runners

### LINKED ISSUE #848: Automatic integration tests triggers are broken
We use the `integration-*.yml` workflows to trigger integration tests outside this repository.

These workflows no longer work because of security hardening.

This issue is a placeholder to link to and to track re-enabling these workflows.

pietern:
Fixes #848.

### DIFF
--- a/.github/workflows/external-message.yml
+++ b/.github/workflows/external-message.yml
@@ -13,7 +13,10 @@ on:
 
 jobs:
   comment-on-pr:
-    runs-on: ubuntu-latest
+    runs-on:
+      group: databricks-deco-testing-runner-group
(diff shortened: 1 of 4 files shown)


.github/workflows/external-message.yml
```diff
@@ -13,7 +13,10 @@ on:
 
 jobs:
   comment-on-pr:
-    runs-on: ubuntu-latest
+    runs-on:
```
hectorcast-db, line 16: This may be better to run in the databricks (non deco) worker, but it's currently not available.
The deco one should only be

hectorcast-db (approved):

PR #852, opened 2025-01-06
[ML-48691] Adding http_request to serving to call /external-function

sunishsheth2009:
[ML-48691] Adding http_request to serving to call /external-function

## Why are these changes needed?
The changes are needed to make it easy for the AI agent authors to create tools that can make request to external services by invoking the /external-function API where the secrets are stored in UC Connections.

The Mixin is added so that it is easier for developers to just pass in json objects rather than stringified data similar to [request.request api](https://requests.readthedocs.io/en/latest/api/) from python 

## How is this tested?
- Not tested yet. Once the approach is clear, will test it in a notebook


databricks/sdk/mixins/open_ai_client.py
```diff
@@ -50,3 +55,47 @@ def get_langchain_chat_open_ai_client(self, model):
...
+            body["json"] = js.dumps(json)
+        if method is not None:
+            body["method"] = method.value
+        if params is not None:
+            body["params"] = js.dumps(params)
+        if path is not None:
+            body["path"] = path
+        headers = {"Accept": "application/json", "Content-Type": "application/json", }
+
+        res = self._api.do("POST", "/external-function", body=body, headers=headers)
```
renaudhartert-db, line 100: The way the generated function is actually making the call might change. We should implement this extension by using the generated function rather than using the API client directly.

```
response = super().http_request(...)
```

databricks/sdk/mixins/open_ai_client.py
```diff
@@ -50,3 +55,47 @@ def get_langchain_chat_open_ai_client(self, model):
...
+        :param headers: Dict[str,str] (optional)
+          Additional headers for the request. If not provided, only auth headers from connections would be
+          passed.
+        :param json: Dict[str,str] (optional)
+          JSON payload for the request.
+        :param params: Dict[str,str] (optional)
+          Query parameters for the request.
+
+        :returns: :class:`ExternalFunctionResponse`
+        """
```
renaudhartert-db, line 84: Could we indicate clearly that the function is experimental and the interface might change?

sunishsheth2009:
Code merged here: https://github.com/databricks/databricks-sdk-py/pull/857

PR #870, opened 2025-01-22
[Fix] Fixing http_request open_ai_client mixin

sunishsheth2009:
- [Fix] Fixing http_request open_ai_client mixin

## How is this tested?
https://dbc-436b86a5-6d5b.dev.databricks.com/editor/notebooks/2465247895264388?o=1593823105423000#command/7927901211745840

```
from databricks.sdk.service.serving import ExternalFunctionRequestHttpMethod

WorkspaceClient().serving_endpoints.http_request(
    conn="test_sql_slack",
    method=ExternalFunctionRequestHttpMethod.POST,
    path="/api/chat.postMessage",
    json={"channel": "C07MB25Q6H3", "text": "Hello!"}
)
```

Tried this and it works

### DIFF
--- a/databricks/sdk/mixins/open_ai_client.py
+++ b/databricks/sdk/mixins/open_ai_client.py
@@ -1,12 +1,32 @@
 import json as js
-from typing import Dict, Optional
+from dataclasses import asdict, dataclass
+from typing import Any, Dict, Optional
 
+import databricks
 from databricks.sdk.service.serving import (ExternalFunctionRequestHttpMethod,
                                             ExternalFunctionResponse,
                                             ServingEndpointsAPI)
 
 
 class ServingEndpointsExt(ServingEndpointsAPI):
+
+    @dataclass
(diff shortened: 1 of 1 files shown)


databricks/sdk/mixins/open_ai_client.py
```diff
@@ -82,10 +102,15 @@ def http_request(self,
...
         """
 
-        return super.http_request(connection_name=conn,
-                                  method=method,
-                                  path=path,
-                                  headers=js.dumps(headers),
-                                  json=js.dumps(json),
-                                  params=js.dumps(params),
-                                  )
+        databricks.sdk.service.serving.ExternalFunctionResponse = (
```
hectorcast-db, line 105: AFAIK, this will change ALL calls to the original http_request. We must avoid these types of side effects. Even if we revert it after the call, it will cause issues in multi threaded environments.

PR #877, opened 2025-01-29
[Fix] Exclude localhost from hitting proxies for metadata service requests

ilia-db:

Exclude localhost from hitting proxies for metadata service requests

`requests` package doesn't do it by default.

Relevant issue: https://github.com/databricks/databricks-vscode/issues/916#issuecomment-2619382290

## How is this tested?

Tested manually by running this code against real metadata service of the logged in databricks vscode extension:
```python
import requests
import os

os.environ["HTTP_PROXY"] = "http://test.com"

resp = requests.get("http://127.0.0.1:53468/redacted",
                    timeout=10000,
                    headers={
                        "X-Databricks-Metadata-Version": "1",
                        "X-Databricks-Host": "https://redacted.databricks.com/"
                    }, proxies={"no_proxy": "localhost,127.0.0.1"})

print(resp.text)
``` 
The code fails without the no_proxy option


databricks/sdk/credentials_provider.py
```diff
@@ -676,12 +676,17 @@ def __init__(self, cfg: 'Config'):
...
+            self.url,
+            timeout=self._metadata_service_timeout,
+            headers={
+                self.METADATA_SERVICE_VERSION_HEADER: self.METADATA_SERVICE_VERSION,
+                self.METADATA_SERVICE_HOST_HEADER: self.host
+            },
+            proxies={
+                # Metadata URL usually points to localhost, and the requests package doesn't exclude localhost from proxies by default
+                "no_proxy": "localhost,127.0.0.1"
+            })
```
renaudhartert-db, line 689:
```suggestion
            proxies={
                # Explicitly exclude the localhost from being proxied. This is necessary
                # for Metadata URLs which typically point to localhost.
                "no_proxy": "localhost,127.0.0.1"
            })
```

renaudhartert-db (approved):

PR #879, opened 2025-01-29
[Internal] Add unit tests for retriable requests

ksafonov-db:

Improving tests for retriable requests, ensuring we're covering retries from error response and connection exceptions.
Now we have one unified test which is parameterized by:
1) data passed to be uploaded
2) type of failure (retriable response or retriable exception)

## How is this tested?

This PR only changes tests.

### DIFF
--- a/tests/test_base_client.py
+++ b/tests/test_base_client.py
@@ -1,10 +1,11 @@
 import io
 import random
 from http.server import BaseHTTPRequestHandler
-from typing import Iterator, List
+from typing import Callable, Iterator, List, Optional
 from unittest.mock import Mock
 
 import pytest
+from requests import PreparedRequest, Response, Timeout
 
 from databricks.sdk import errors, useragent
 from databricks.sdk._base_client import (_BaseClient, _RawResponse,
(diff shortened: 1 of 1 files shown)


tests/test_base_client.py
```diff
@@ -357,91 +358,154 @@ def tell(self):
...
+                 expected_result: bytes):
+        self._data_provider = data_provider
+        self._offset = offset
+        self._expected_result = expected_result
+        self._expected_exception = expected_exception
 
-        content_length = int(h.headers.get('Content-Length', 0))
-        if content_length > 0:
-            received_data.append(h.rfile.read(content_length))
+    def get_data(self):
```
ksafonov-db, line 370: Test cases are reused, so we need to construct a fresh data object every time.

tests/test_base_client.py
```diff
@@ -357,91 +358,154 @@ def tell(self):
...
+            with pytest.raises(DatabricksError):
+                do()
+        else:
+            expected_attempts_made = failure_count + 1
+            do()
+
+        assert received_requests == [test_case._expected_result for _ in range(expected_attempts_made)]
+
+
+class MockSession:
```
ksafonov-db, line 440: This mock session reads all the input stream and then throws retriable exception.

renaudhartert-db (approved):

PR #882, opened 2025-01-31
[Feature] Introduce new Credential Strategies for Agents

aravind-segu:

This PR introduces two new credential strategies for Agents, (AgentEmbeddedCredentials, AgentUserCredentials). 

Agents currently use the databricks.sdk in order to interact with databricks resources. However the authentication method for these resources is a little unique where we store the token for the authentication in a Credential File on the Kubernetes Container. Therefore in the past we added the Model Serving Credential Strategy to the defaultCredentials list to read this file. 

Now we want to introduce a new authentication where the user's token is instead stored in a thread local variable. Agent users will initialize clients as follows:

```
from databricks.sdk.credentials_provider import ModelServingUserCredentials

invokers_client = WorkspaceClient(credential_strategy = ModelServingUserCredentials())
definers_client = WorkspaceClient()

```

Then the users can use the invoker_client to interact with resources with the invokers token or the definers_client to interact with resources using the old method of authentication. 

Additionally as the users will be using these clients to test their code locally in Databricks Notebooks, if the code is not being run on model serving environments, users need to be able to authenticate using the DefaultCredential strategies. 

More details: https://docs.google.com/document/d/14qLVjyxIAk581w287TWElstIeh8-DR30ab9Z6B_Vydg/edit?usp=sharing

## How is this tested?

Added unit tests

### DIFF
--- a/databricks/sdk/credentials_provider.py
+++ b/databricks/sdk/credentials_provider.py
@@ -314,11 +314,12 @@ def github_oidc_azure(cfg: 'Config') -> Optional[CredentialsProvider]:
         # detect Azure AD Tenant ID if it's not specified directly
         token_endpoint = cfg.oidc_endpoints.token_endpoint
         cfg.azure_tenant_id = token_endpoint.replace(aad_endpoint, '').split('/')[0]
-    inner = ClientCredentials(client_id=cfg.azure_client_id,
-                              client_secret="", # we have no (rotatable) secrets in OIDC flow
-                              token_url=f"{aad_endpoint}{cfg.azure_tenant_id}/oauth2/token",
-                              endpoint_params=params,
-                              use_params=True)
+    inner = ClientCredentials(
+        client_id=cfg.azure_client_id,
+        client_secret="", # we have no (rotatable) secrets in OIDC flow
+        token_url=f"{aad_endpoint}{cfg.azure_tenant_id}/oauth2/token",
+        endpoint_params=params,
+        use_params=True)
 
     def refreshed_headers() -> Dict[str, str]:
         token = inner.token()
--- a/databricks/sdk/credentials_provider.py
+++ b/databricks/sdk/credentials_provider.py
@@ -314,11 +314,12 @@ def github_oidc_azure(cfg: 'Config') -> Optional[CredentialsProvider]:
         # detect Azure AD Tenant ID if it's not specified directly
         token_endpoint = cfg.oidc_endpoints.token_endpoint
         cfg.azure_tenant_id = token_endpoint.replace(aad_endpoint, '').split('/')[0]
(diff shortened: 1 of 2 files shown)


databricks/sdk/credentials_provider.py
```diff
@@ -758,25 +763,38 @@ def get_model_dependency_oauth_token(self, should_retry=True) -> str:
...
+        if not ModelServingAuthProvider.should_fetch_model_serving_environment_oauth():
             return None
 
         # read from DB_MODEL_SERVING_HOST_ENV_VAR if available otherwise MODEL_SERVING_HOST_ENV_VAR
         host = os.environ.get("DATABRICKS_MODEL_SERVING_HOST_URL") or os.environ.get(
             "DB_MODEL_SERVING_HOST_URL")
-        token = self.get_model_dependency_oauth_token()
+        token = self.get_model_dependency_oauth_token(
+        ) if self.credential_type == ModelServingAuthProvider.EMBEDDED_CREDENTIALS else self.get_invokers_token(
+        )
```
renaudhartert-db, line 787: [nit] This is a little hard to read. Maybe something like the following?

```suggestion
        if self.credential_type == ModelServingAuthProvider.EMBEDDED_CREDENTIALS:
            return (host, self.get_model_dependency_oauth_token())
        else:
            return (host, self.get_invokers_token())
```

databricks/sdk/credentials_provider.py
```diff
@@ -758,25 +763,38 @@ def get_model_dependency_oauth_token(self, should_retry=True) -> str:
...
 
         return (host, token)
 
 
-@credentials_strategy('model-serving', [])
-def model_serving_auth(cfg: 'Config') -> Optional[CredentialsProvider]:
+def model_serving_auth_func(cfg: 'Config', credential_type) -> Optional[CredentialsProvider]:
```
renaudhartert-db, line 792: These functions are often named "visitor" in our codebase, let's keep it consistent. 

```suggestion
def model_serving_auth_visitor(cfg: 'Config', credential_type) -> Optional[CredentialsProvider]:
```

databricks/sdk/credentials_provider.py
```diff
@@ -840,3 +863,27 @@ def __call__(self, cfg: 'Config') -> CredentialsProvider:
         raise ValueError(
             f'cannot configure default credentials, please check {auth_flow_url} to configure credentials for your preferred authentication method.'
         )
+
+
+class ModelServingUserCredentials(CredentialsStrategy):
```
renaudhartert-db, line 868: Would be good to describe the class, what it does, and it what situation it is meant to be used. Something similar to what is currently in your PR description.

renaudhartert-db (approved):
LGTM assuming resolution of my last comment. Thanks @aravind-segu!

PR #893, opened 2025-02-19
[Internal] Implement async token refresh

hectorcast-db:

This PR is a step towards enabling asynchronous refreshes of data plane tokens.
This PR updates the existing `Refreshable` abstract token class to support async token refresh.

Note: async refreshes are disabled at the moment and will be enabled in a follow-up PR.

## How is this tested?

Added unit tests.

## Changelog
The changelog entry will be added when the feature is enabled.

NO_CHANGELOG=true

### DIFF
--- a/databricks/sdk/oauth.py
+++ b/databricks/sdk/oauth.py
@@ -228,16 +228,19 @@ class Refreshable(TokenSource):
         self._disable_async = disable_async
         # Lock
         self._lock = threading.Lock()
-        # Non Thread safe properties. Protected by the lock above.
+        # Non Thread safe properties. They should be accessed only when protected by the lock above.
         self._token = token
         self._is_refreshing = False
         self._refresh_err = False
 
+    # This is the main entry point for the Token. Do not access the token
+    # using any of the internal functions.
     def token(self) -> Token:
         """Returns a valid token, blocking if async refresh is disabled."""
-        if self._disable_async:
-            return self._blocking_token()
-        return self._async_token()
+        with self._lock:
+            if self._disable_async:
+                return self._blocking_token()
+            return self._async_token()
 
     def _async_token(self) -> Token:
         """
@@ -245,9 +248,8 @@ class Refreshable(TokenSource):
         If the token is stale, triggers an asynchronous refresh.
         If the token is expired, refreshes it synchronously, blocking until the refresh is complete.
         """
-        with self._lock:
-            state = Refreshable._token_state(self._token, self._stale_duration)
-            token = self._token
+        state = self._token_state()
+        token = self._token
 
         if state == _TokenState.FRESH:
             return token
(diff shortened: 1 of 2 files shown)


databricks/sdk/oauth.py
```diff
@@ -187,21 +189,117 @@ def retrieve_token(client_id,
...
+    _EXECUTOR_LOCK = threading.Lock()
+    _DEFAULT_STALE_DURATION = timedelta(minutes=3)
+
+    @classmethod
+    def _get_executor(cls):
+        """Lazy initialization of the ThreadPoolExecutor."""
+        if cls._EXECUTOR is None:
+            with cls._EXECUTOR_LOCK:
+                if cls._EXECUTOR is None:
+                    cls._EXECUTOR = ThreadPoolExecutor(max_workers=10)
```
renaudhartert-db, line 218: Maybe a small note to explain why not "1"?

renaudhartert-db (commented):
Looks good to me overall.

As discussed offline, the first version had a problem with the ThreadPoolExecutor being created when the class was loaded. This is fixed now.

databricks/sdk/oauth.py
```diff
@@ -187,21 +189,118 @@ def retrieve_token(client_id,
...
+    def __init__(self,
+                 token: Token = None,
+                 disable_async: bool = True,
+                 stale_duration: timedelta = _DEFAULT_STALE_DURATION):
+        self._lock = threading.Lock()
         self._token = token
+        self._stale_duration = stale_duration
+        self._disable_async = disable_async
+        self._is_refreshing = False
+        self._refresh_err = False
```
mgyucht, line 231: Is there a way to indicate which properties of this class must be accessed through the lock?

databricks/sdk/oauth.py
```diff
@@ -187,21 +189,131 @@ def retrieve_token(client_id,
...
+    def _async_token(self) -> Token:
+        """
+        Returns a token.
+        If the token is stale, triggers an asynchronous refresh.
+        If the token is expired, refreshes it synchronously, blocking until the refresh is complete.
+        """
+        token_state = self._token_state()
+
+        with self._lock:
+            token = self._token
```
mgyucht, line 251: Because we release the lock here, the token on line 251 could be different than the one used in `_token_state()`. If we convert `_token_state()` to something that returns the token and its state, we can avoid that problem. Thoughts?

databricks/sdk/oauth.py
```diff
@@ -187,21 +189,131 @@ def retrieve_token(client_id,
...
+
+            with self._lock:
+                if new_token is not None:
+                    self._token = new_token
+                else:
+                    self._refresh_err = True
+                self._is_refreshing = False
+
+        with self._lock:
+            if not self._is_refreshing and not self._refresh_err:
```
mgyucht, line 314: I think this is safe but might fetch more tokens than necessary. Like you mentioned offline, the first thread to acquire this lock will start the refresh process, and then all other threads that acquire the lock while refresh is happening will skip. However, once refresh is complete, other threads that acquire the lock (e.g. that checked that the token was stale and came down this pathway while the refresh was ongoing, then only acquired this lock after refresh was completed) will unnecessarily refresh the token again. So I think we should also check the token state again here: if it is not stale, skip the refresh.

databricks/sdk/oauth.py
```diff
@@ -187,21 +189,136 @@ def retrieve_token(client_id,
...
+                    self._token = new_token
+                else:
+                    self._refresh_err = True
+                self._is_refreshing = False
+
+        with self._lock:
+            state = Refreshable._token_state(self._token, self._stale_duration)
+            # The token may have been refreshed by another thread.
+            if state == _TokenState.FRESH:
+                return
```
renaudhartert-db, line 318: While the comment is 100% correct, I don't think this is the right place to optimize. Rather, I believe it would make more sense to execute the async refresh as part of the lock acquisition that happens in line 248.

renaudhartert-db (changes requested):

renaudhartert-db (approved):

mgyucht (approved):

PR #896, opened 2025-02-21
[Internal] Update Jobs list function to support paginated responses

hectorcast-db:
Introduces logic in the extension for jobs ListJobs call. The extended logic accounts for the new response format of API 2.2. API 2.1 format returns all tasks and job_cluster for each job in the jobs list. API 2.2 format truncates tasks and job_cluster to 100 elements. The extended ListJobs logic calls GetJob for each job in the list to populate the full list of tasks and job_clusters.

The code consumes generator that is returned from `super().list` and it produces generator as well.

## How is this tested?
Unit tests and manual tests. Manual tests were done in two modes: using API 2.2 and using API 2.1. So this code is code is compatible with both API versions.

## Notes
Copy of this other PR
https://github.com/databricks/databricks-sdk-py/pull/889

The formatter is inconsistent and the original author could make it work.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Documentation
 
 ### Internal Changes
+* Update Jobs ListJobs API to support paginated responses ([#896](https://github.com/databricks/databricks-sdk-py/pull/896))
 * Introduce automated tagging ([#888](https://github.com/databricks/databricks-sdk-py/pull/888))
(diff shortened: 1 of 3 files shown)


hectorcast-db:
Copy of this other PR
https://github.com/databricks/databricks-sdk-py/pull/889

The make fmt is inconsistent and the original reviewer could not have it work

renaudhartert-db (approved):

PR #897, opened 2025-02-24
[Internal] Add DataPlane token source

hectorcast-db:

Add DataPlane token source using the new async refresh mechanism. 
This first version has async disabled.

## How is this tested?
Added unit tests

NO_CHANGELOG=true

This is not yet used.


databricks/sdk/data_plane.py
```diff
@@ -1,9 +1,69 @@
...
+        Get a token for a specific DataPlane endpoint.
+        :param endpoint: endpoint URL for which to get a token
+        :param auth_details:  authorization details used to generate the token
+        :return: a token for the specified endpoint
+        """
+        key = f"{endpoint}:{auth_details}"
+        token_source = self._token_sources.get(key)
+        if not token_source:
+            token_source = DataPlaneEndpointTokenSource(self._cfg, auth_details, self._disable_async)
+            self._token_sources[key] = token_source
```
renaudhartert-db, line 38: Should this be made thread safe?

databricks/sdk/data_plane.py
```diff
@@ -1,9 +1,69 @@
...
+
+
+class DataPlaneTokenSource:
+    """
+    EXPERIMENTAL Manages token sources for multiple DataPlane endpoints.
+    """
+
+    # TODO: Enable async once its stable. @oauth_credentials_provider must also have async enabled.
+    def __init__(self, cfg: config.Config, disable_async: Optional[bool] = True):
+        self._cfg = cfg
```
renaudhartert-db, line 23: Could we decouple this from `config.Config`? This class is ultimately meant to move in the `auth` library which will not know about `config.Config`.

hectorcast-db: Done

renaudhartert-db (approved):

PR #900, opened 2025-02-26
[Internal] Switch code formatter to Black

renaudhartert-db:

This PR switches the code formatter from `yapf` to `black`. 

The rationale is that we want a code formatter that (i) leaves as little space as possible for discussions around formatting personal preferences, and (ii) increases the likelihood of producing code that is consistently formatted across the python open-source ecosystem. 

At this point in time, we believe that `black` does a better job than `yapf` and other alternatives to address these two problems. 

## How is this tested?

No additional tests.

NO_CHANGELOG=true


denik:
Why not ruff? It is compatible with black and comes with a lot of lint rules. Much faster as well.

renaudhartert-db:
> Why not ruff? It is compatible with black and comes with a lot of lint rules. Much faster as well.

Speed and linting are not problems we're trying to address at this point in time. It is possible that we will converge towards `Ruff` once we're done experimenting and have identified the tools we want to rely on. For now, we believe that moving to `Black` is a strict improvement of the status quo.

PR #902, opened 2025-02-26
[Feature] Large file uploads (901)

ksafonov-db:

Supporting large file uploads using Files API.

NO_CHANGELOG=true

## How is this tested?

Unit tests.


databricks/sdk/mixins/files.py
```diff
@@ -713,11 +727,560 @@ def download(self, file_path: str) -> DownloadResponse:
...
+
+        def fill_buffer():
+            bytes_to_read = max(0, self._config.multipart_upload_chunk_size - len(buffer))
+            if bytes_to_read > 0:
+                next_buf = input_stream.read(bytes_to_read)
+                new_buffer = buffer + next_buf
+                return new_buffer
+            else:
+                # we have already buffered enough data
+                return buffer
```
renaudhartert-db, line 810: I think this function is a little confusion as it looks functional and procedural at the same time. I think it would help to declare it as a standalone static function that returns the next buffer. Something like this:

```py
def _read_next_buffer(
    input_stream: BinaryIO, 
    previous_buffer: bytes, 
    chunk_size: int=Config.MULTIPART_UPLOAD_CHUNK_SIZE,
) -> bytes:
    """
    Returns the next buffer to be processed.

    The size of the next buffer can be more or less than the given chunk_size. If it is
    less, it means that the all the input_stream has been read and that 
    
    The function avoids reading new bytes from the input_stream if previous_buffer
    already contains chunk_size bytes to be read.
    """

    bytes_to_read = max(0, Config.MULTIPART_UPLOAD_CHUNK_SIZE - len(previous_buffer))

    if bytes_to_read <= 0:
        return previous_buffer # buffer already contains a full chunk

    next_bytes = input_stream.read(bytes_to_read)
    new_buffer = previous_buffer + next_bytes
    return new_buffer
```

All that behind said, I think it would be better to abstract all this in a class that would hide the buffering logic and expose simple interface that read the next bytes. This can come as a follow-up to this PR though.

databricks/sdk/mixins/files.py
```diff
@@ -713,11 +727,560 @@ def download(self, file_path: str) -> DownloadResponse:
...
+
+            # Requesting URLs for the same set of parts is an idempotent operation, safe to retry.
+            # _api.do() does retry
+            upload_part_urls_response = self._api.do(
+                "POST", "/api/2.0/fs/create-upload-part-urls", headers=headers, body=body
+            )
+            # no need to check response status, _api.do() will throw exception on failure
+
+            upload_part_urls = upload_part_urls_response.get("upload_part_urls", [])
+            if not len(upload_part_urls):
```
renaudhartert-db, line 843: Let's aim for clarity for developers who are not use to Python idioms.

```suggestion
            if len(upload_part_urls) == 0:
```

databricks/sdk/mixins/files.py
```diff
@@ -713,11 +727,560 @@ def download(self, file_path: str) -> DownloadResponse:
...
+                raise ValueError(f"Unexpected server response: {upload_part_urls_response}")
+
+            for upload_part_url in upload_part_urls:
+                buffer = fill_buffer()
+                actual_buffer_length = len(buffer)
+                if not actual_buffer_length:
```
renaudhartert-db, line 849: Same as above, while maybe more "pythonic", this adds ambiguity.

renaudhartert-db (approved):
LGTM, I think we'll have to do a little more work on the code style (e.g. use types that are compatible with python 3.7) but this should not be a blocker to start getting customer feedback on this experimental function. Test coverage is good and I'm confident that the code is doing what it has been designed to do.

PR #914, opened 2025-03-05
[Internal] Fix some incorrect type hints

renaudhartert-db:

This PR is a no-op that fixes some incorrect type hints.

## How is this tested?

Nothing more than unit tests and integration tests. 

NO_CHANGELOG=true


databricks/sdk/environments.py
```diff
@@ -113,7 +113,7 @@ def azure_active_directory_endpoint(self) -> Optional[str]:
 ]
 
 
-def get_environment_for_hostname(hostname: str) -> DatabricksEnvironment:
+def get_environment_for_hostname(hostname: Optional[str]) -> DatabricksEnvironment:
```
parthban-db, line 116: Why is this marked as optional?

renaudhartert-db: Good question. Other methods are passing optional string to it and the function is designed to handle the None value already.

databricks/sdk/_base_client.py
```diff
@@ -403,10 +403,10 @@ def truncate(self, __size: Union[int, None] = ...) -> int:
     def writable(self) -> bool:
         return False
 
-    def write(self, s: Union[bytes, bytearray]) -> int:
+    def write(self, s: Union[bytes, bytearray]) -> int:  # type: ignore
         raise NotImplementedError()
```
parthban-db, line 407: Why #type: ignore here? It seems to be correctly typed.

renaudhartert-db: Same as above.

parthban-db (approved):

PR #919, opened 2025-03-07
[Internal] Single-shot upload reports content length if known

ksafonov-db:

When performing single-shot upload, determine content length (via utility method from Python requests library), and if length is known, fill the `Content-Length` HTTP header. Then server will respect this header to validate request body size against.

## How is this tested?

Added unit test.

NO_CHANGELOG=true

### DIFF
--- a/tests/test_files.py
+++ b/tests/test_files.py
@@ -1940,7 +1940,14 @@ def create_file_stream(length: int) -> io.IOBase:
         f.write(os.urandom(length))
 
     stream = open(temp_file, "rb")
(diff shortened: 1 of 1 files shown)


tests/test_files.py
```diff
@@ -1279,8 +1279,6 @@ def run(self, config: Config):
 
         session = requests.Session()
         with requests_mock.Mocker(session=session) as session_mock:
-            session_mock.get(f"http://localhost/api/2.0/fs/files{MultipartUploadTestCase.path}", status_code=200)
```
ksafonov-db, line 1282: cleanup

PR #923, opened 2025-03-09
Add Service Modules Imports to __init__.py

parthban-db:

What - This PR updates the `__init__.py` file in the `databricks/sdk/service` directory to import and re-export all service modules. 

Why - This change will make the modules accessible directly from the `databricks.sdk.service` namespace without requiring users to import each service module individually. We access services through `databricks.sdk.service` while creating workspace and account client but we don't export it.
```
        self._access_control = service.iam.AccountAccessControlAPI(self._api_client)
        self._billable_usage = service.billing.BillableUsageAPI(self._api_client)
```

## How is this tested?
Existing Tests.
NO_CHANGELOG=true

### DIFF
--- a/databricks/sdk/service/__init__.py
+++ b/databricks/sdk/service/__init__.py
@@ -1,31 +1,16 @@
 # Import all service modules
-from databricks.sdk.service import apps
-from databricks.sdk.service import billing
-from databricks.sdk.service import catalog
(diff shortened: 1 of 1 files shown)


renaudhartert-db:
Discussed offline, we will address this at the code generation level.

PR #731, opened 2024-08-19
Add BROWSE option in Privilege Enum

jeremyrodrigues:
## Changes
Simple fix

Issue #639, opened 2024-05-08
[ISSUE] Missing `BROWSE` Privilege on a `Catalog` object


green2k:
**Description**
There's the `BROWSE` value missing in `databricks.sdk.service.catalog.Privilege` enum.

**Reproduction**
A minimal code sample demonstrating the bug.
```python
from databricks.sdk.service.catalog import Privilege
print(Privilege.BROWSE)
```

**Expected behavior**
Prints `BROWSE` enum value

**Is it a regression?**
I haven't found any version of the SDK, where this works.

**Debug Logs**
n/a

**Other Information**
 - OS: `Ubuntu`
 - Version: `0.27.0`

**Additional context**

![image](https://github.com/databricks/databricks-sdk-py/assets/6991578/63ef520b-1a37-4e75-96da-ac65b2112fa0)

ilyakisleika:
fix: https://github.com/databricks/databricks-sdk-py/pull/731

tsktsktsk123:
I believe this can be closed now as it has been present since v0.49 
https://github.com/databricks/databricks-sdk-py/releases/tag/v0.49.0

PR #929, opened 2025-03-19
Added BROWSE privilege

Tim-Hoare:
## Changes

Added [BROWSE](https://learn.microsoft.com/en-us/azure/databricks/data-governance/unity-catalog/manage-privileges/privileges#browse) privilege to `Privilege` enum. 

This is a duplicate of #731


ilyakisleika:
no longer required, already fixed, see https://github.com/databricks/databricks-sdk-py/issues/639#issuecomment-2817067498

PR #933, opened 2025-03-21
Support Databricks Workload Identity Federation for GitHub tokens

hectorcast-db:

This PR adds support for Databricks Workload Identity Federation using GitHub tokens. This allows users to use WIF from their GitHub Workflows and authenticate their workloads without long lived secrets. 

This new credentials strategy is added to the DefaultCredentialsStrategy after the other Databricks Credentials Strategy and before cloud specific authentication methods.
WIF credentials uses a subset of configuration values of other Databricks authentication methods. By being added after them it ensures that WIF is not used when  other Databricks authentication methods are configured.
WIF uses the Databricks client id, which is not used by cloud specific authentication methods. Therefore, it will not be used when cloud specific authentication methods are configured.

## How is this tested?
Added tests.


renaudhartert-db (approved):
LGTM as soon as we have agreement on the auth_type name.

PR #941, opened 2025-03-28
Arrange services in a per package sdk approach

parthban-db:

This PR moves all the service packages into separate packages and isolates them from each other. It also removes the workspace and account clients, which are replaced by the per-service clients. 
Note: waiters and data plane services are removed and need to be enabled again.
## How is this tested?

- Modified existing tests to use service clients instead of workspace and account clients.
- Few tests are commented out, like tests involving waiters.
- WorkspaceClient, AccountClient Test creation tests are deleted.

NO_CHANGELOG=true


databricks/sdk/__init__.py
```diff
@@ -9,104 +9,8 @@
...
-                                        RedashConfigAPI, StatementExecutionAPI,
-                                        WarehousesAPI)
-from databricks.sdk.service.vectorsearch import (VectorSearchEndpointsAPI,
-                                                 VectorSearchIndexesAPI)
-from databricks.sdk.service.workspace import (GitCredentialsAPI, ReposAPI,
-                                              SecretsAPI, WorkspaceAPI)
+from databricks.sdk.files.v2.files import FilesAPI
+from databricks.sdk.files.v2.mixin import DbfsExt, FilesExt
 
 _LOG = logging.getLogger(__name__)
```
hectorcast-db, line 15: Do we still need these files related functions here?

parthban-db: Hmm, I think we can remove it. I am wondering why they are not removed by black. I will remove these and see what happens to the tests. If they fail, I will keep it as it is for now and try to remove it in the following PRs.

renaudhartert-db (approved):

PR #945, opened 2025-04-07
Added missing privilege EXTERNAL USE SCHEMA

jijosg:

When trying to assign the grant EXTERNAL USE SCHEMA to a schema , it doesn't pick up from the existing list of grants.

- **WHAT** Added a new privilege EXTERNAL USE SCHEMA in the Privilege class in databricks/sdk/service/catalog.py
- **WHY** are these changes needed? This should provide the context that the 
  reader might be missing. For example, were there any decisions behind the 
  change that are not reflected in the code itself? 

The EXTERNAL USE SCHEMA is a new feature that was introduced to be able to access data registered in Unity Catalog from clients external to Databricks. This is currently missing in the code.

## How is this tested?
Not tested

NO_CHANGELOG=true


jijosg:
Looks like this repo is not being actively maintained

alexkot1394:
Hi folks! I'm working on a service that provisions Kafka Connect jobs and I'm adding an iceberg databricks sink. In order to use the Iceberg rest api with Databricks we need to grant our SPs EXTERNAL USE SCHEMA  ([link to doc](https://docs.databricks.com/aws/en/external-access/admin)). I'm trying to add this grant to our presets for our Data Product management tool that uses this SDK. I'm running into an issue where EXTERNAL USE SCHEMA is not one of the privilege types in the databricks python SDK. We need this PR to be merged and released to resolve our issue.

jijosg:
Fixed in https://github.com/databricks/databricks-sdk-py/releases/tag/v0.67.0

PR #949, opened 2025-04-11
Adding an example to show how to use api_client to make generic api calls. 

harrydu-db:

I am adding one example under `examples`.  It is to demonstrate how to use `api_client` for making generic REST API calls. This serves as an example for users who need to interact with REST endpoints that are not yet wrapped by the Python SDK.

## How is this tested?

 - [x] make test run locally
 - [x] make fmt applied

### DIFF
--- a/examples/generic_restapi.py
+++ b/examples/generic_restapi.py
@@ -1,21 +1,20 @@
 from databricks.sdk import WorkspaceClient
-import logging
-logging.basicConfig(level=logging.DEBUG)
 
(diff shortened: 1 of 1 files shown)


renaudhartert-db:
The API client is actually not meant to be used directly and will likely become private in the future. The Databricks SDK aims to offer clients for all public Databricks API. If one API is not included, then it's probably due to an upstream issue that we need to resolve internally. Please feel free to reach out internally for more details.

PR #962, opened 2025-05-07
[DONT REVIEW, JUST FOR TESTING]Set `stacklevel` for `warnings.warn`

harupy:

Provide the readers and reviewers with the information they need to understand
this PR in a comprehensive manner. 

- **WHAT** changes are being made in the PR? This should be a summary of the 
  major changes to allow the reader to quickly understand the PR without having
  to look at the code. 
- **WHY** are these changes needed? This should provide the context that the 
  reader might be missing. For example, were there any decisions behind the 
  change that are not reflected in the code itself? 

The “why part” is the most important of the two as it usually cannot be 
inferred from the code itself. A well-written PR description will help future
developers (including your future self) to know how to interact and update your
code.

## How is this tested?

**ALWAYS ANSWER THIS QUESTION:** Answer with "N/A" if tests are not applicable
to your PR (e.g. if the PR only modifies comments). Do not be afraid of 
answering "Not tested" if the PR has not been tested. Being clear about what 
has been done and not done provides important context to the reviewers.

### DIFF
--- a/databricks/sdk/errors/base.py
+++ b/databricks/sdk/errors/base.py
@@ -84,6 +84,7 @@ class DatabricksError(IOError):
 
         # Retry-after is deprecated.
         if retry_after_secs:
+            raise Exception("who uses retry_after_secs?")
             warnings.warn(
                 "The 'retry_after_secs' parameter of DatabricksError is deprecated and will be removed in a future version.",
                 stacklevel=2,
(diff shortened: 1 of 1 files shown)


databricks/sdk/errors/base.py
```diff
@@ -85,7 +85,8 @@ def __init__(
         # Retry-after is deprecated.
         if retry_after_secs:
             warnings.warn(
```
harupy, line 87:
```suggestion
            raise Exception("who uses retry_after_secs?")
            warnings.warn(
```

PR #975, opened 2025-05-21
Add custom Duration and Timestamp classes with nanosecond support

hectorcast-db:
Add custom Duration and Timestamp classes with nanosecond support

## How is this tested?
Added tests

### DIFF
--- a/databricks/sdk/common.py
+++ b/databricks/sdk/common.py
@@ -1,0 +1,313 @@
+"""Common types for the Databricks SDK.
+
+This module provides common types used by different APIs.
+"""
+
+from __future__ import annotations
+
+import logging
+import re
+from datetime import datetime, timedelta, timezone
+
+_LOG = logging.getLogger("databricks.sdk")
+
+# Python datetime library does not have nanoseconds precision. These classes below are used to work around this limitation.
+
+
+class Duration:
+    """Represents a duration with nanosecond precision.
+
+    This class provides nanosecond precision for durations, which is not supported
+    by Python's standard datetime.timedelta.
+
+    Attributes:
+        seconds (int): Number of seconds in the duration
+        nanoseconds (int): Number of nanoseconds (0-999999999)
+    """
+
+    def __init__(self, seconds: int = 0, nanoseconds: int = 0) -> None:
+        """Initialize a Duration with seconds and nanoseconds.
+
+        Args:
+            seconds: Number of seconds
+            nanoseconds: Number of nanoseconds (0-999999999)
+
+        Raises:
+            TypeError: If seconds or nanoseconds are not integers
+            ValueError: If nanoseconds is not between 0 and 999999999
+        """
+        if not isinstance(seconds, int):
+            raise TypeError("seconds must be an integer")
+        if not isinstance(nanoseconds, int):
+            raise TypeError("nanoseconds must be an integer")
+        if nanoseconds < 0 or nanoseconds >= 1_000_000_000:
+            raise ValueError("nanoseconds must be between 0 and 999999999")
+
+        self.seconds = seconds
+        self.nanoseconds = nanoseconds
+
+    @classmethod
+    def from_timedelta(cls, td: timedelta) -> "Duration":
+        """Convert a datetime.timedelta to Duration.
+
+        Args:
+            td: The timedelta to convert
+
+        Returns:
+            Duration: A new Duration instance with equivalent time span
+
+        Note:
+            The conversion may lose precision as timedelta only supports microsecond precision
+        """
+        total_seconds = int(td.total_seconds())
+        # Get the microseconds part and convert to nanoseconds
+        microseconds = td.microseconds
+        nanoseconds = microseconds * 1000
+        return cls(seconds=total_seconds, nanoseconds=nanoseconds)
+
+    def to_timedelta(self) -> timedelta:
+        """Convert Duration to datetime.timedelta.
+
+        Returns:
+            timedelta: A new timedelta instance with equivalent time span
+
+        Note:
+            The conversion may lose precision as timedelta only supports microsecond precision
+        """
+        # Convert nanoseconds to microseconds for timedelta
+        microseconds = self.nanoseconds // 1000
+        return timedelta(seconds=self.seconds, microseconds=microseconds)
+
+    def __repr__(self) -> str:
+        """Return a string representation of the Duration.
+
+        Returns:
+            str: String in the format 'Duration(seconds=X, nanoseconds=Y)'
+        """
+        return f"Duration(seconds={self.seconds}, nanoseconds={self.nanoseconds})"
+
+    def __eq__(self, other: object) -> bool:
+        """Compare this Duration with another object for equality.
+
+        Args:
+            other: Object to compare with
+
+        Returns:
+            bool: True if other is a Duration with same seconds and nanoseconds
+        """
+        if not isinstance(other, Duration):
+            return NotImplemented
+        return self.seconds == other.seconds and self.nanoseconds == other.nanoseconds
+
+    @classmethod
+    def parse(cls, duration_str: str) -> "Duration":
+        """Parse a duration string in the format 'Xs' where X is a decimal number.
+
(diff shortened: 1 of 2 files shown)


databricks/sdk/common.py
```diff
@@ -0,0 +1,313 @@
...
+            nanoseconds: Number of nanoseconds (0-999999999)
+
+        Raises:
+            TypeError: If seconds or nanoseconds are not integers
+            ValueError: If nanoseconds is not between 0 and 999999999
+        """
+        if not isinstance(seconds, int):
+            raise TypeError("seconds must be an integer")
+        if not isinstance(nanoseconds, int):
+            raise TypeError("nanoseconds must be an integer")
```
parthban-db, line 42: Why do we need this type check when we explicitly stated that in the function arguments?

databricks/sdk/common.py
```diff
@@ -0,0 +1,313 @@
...
+        if nanoseconds < 0 or nanoseconds >= 1_000_000_000:
+            raise ValueError("nanoseconds must be between 0 and 999999999")
+
+        self.seconds = seconds
+        self.nanoseconds = nanoseconds
+
+    @classmethod
+    def from_timedelta(cls, td: timedelta) -> "Duration":
```
parthban-db, line 50: This is returning "Duration" with quotes. Is this expected?

databricks/sdk/common.py
```diff
@@ -0,0 +1,313 @@
...
+        """Convert a datetime.timedelta to Duration.
+
+        Args:
+            td: The timedelta to convert
+
+        Returns:
+            Duration: A new Duration instance with equivalent time span
+
+        Note:
+            The conversion may lose precision as timedelta only supports microsecond precision
```
parthban-db, line 60: This is incorrect. As we are returning something that has higher precision.

databricks/sdk/common.py
```diff
@@ -0,0 +1,313 @@
...
+        """Compare this Duration with another object for equality.
+
+        Args:
+            other: Object to compare with
+
+        Returns:
+            bool: True if other is a Duration with same seconds and nanoseconds
+        """
+        if not isinstance(other, Duration):
+            return NotImplemented
```
parthban-db, line 99: Why NotImplemented? Why not a failure?

databricks/sdk/common.py
```diff
@@ -0,0 +1,313 @@
...
+        return self.seconds == other.seconds and self.nanoseconds == other.nanoseconds
+
+    @classmethod
+    def parse(cls, duration_str: str) -> "Duration":
```
parthban-db, line 103: ditto: Duration with quotes.

databricks/sdk/common.py
```diff
@@ -0,0 +1,313 @@
...
+
+        Raises:
+            ValueError: If the string format is invalid
+        """
+        if not duration_str.endswith("s"):
+            raise ValueError("Duration string must end with 's'")
+
+        try:
+            # Remove the 's' suffix and convert to float
+            value = float(duration_str[:-1])
```
parthban-db, line 125: Are we sure that we will not lose precision while using floats? Not sure of Python, but pretty sure this will lose precision in other languages like C++.

databricks/sdk/common.py
```diff
@@ -0,0 +1,313 @@
...
+            Duration(seconds=10, nanoseconds=0) -> "10s"
+
+        Returns:
+            String representation of the duration
+        """
+        if self.nanoseconds == 0:
+            return f"{self.seconds}s"
+
+        # Convert to decimal representation
+        total_seconds = self.seconds + (self.nanoseconds / 1_000_000_000)
```
parthban-db, line 149: not sure if this will lose precision or not.

databricks/sdk/common.py
```diff
@@ -0,0 +1,313 @@
...
+            nanos: Nanoseconds (0-999999999)
+
+        Raises:
+            TypeError: If seconds or nanos are not integers
+            ValueError: If nanos is not between 0 and 999999999
+        """
+        if not isinstance(seconds, int):
+            raise TypeError("seconds must be an integer")
+        if not isinstance(nanos, int):
+            raise TypeError("nanos must be an integer")
```
parthban-db, line 185: ditto why int instance check.

tests/test_common.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # Test validation
+    with pytest.raises(TypeError):
+        Duration(seconds="1")  # type: ignore
+    with pytest.raises(TypeError):
+        Duration(nanoseconds="500000000")  # type: ignore
+    with pytest.raises(ValueError):
+        Duration(nanoseconds=-1)
+    with pytest.raises(ValueError):
+        Duration(nanoseconds=1_000_000_000)
```
parthban-db, line 30: I prefer to write this as a parameterized test if possible.

tests/test_common.py
```diff
@@ -0,0 +1,213 @@
...
+    td = timedelta(microseconds=500000)
+    d = Duration.from_timedelta(td)
+    assert d.seconds == 0
+    assert d.nanoseconds == 500000000
+
+    # Test with both
+    td = timedelta(seconds=1, microseconds=500000)
+    d = Duration.from_timedelta(td)
+    assert d.seconds == 1
+    assert d.nanoseconds == 500000000
```
parthban-db, line 51: ditto this and the rest of the tests.

databricks/sdk/common.py
```diff
@@ -0,0 +1,313 @@
...
+        """Compare this Timestamp with another object for equality.
+
+        Args:
+            other: Object to compare with
+
+        Returns:
+            bool: True if other is a Timestamp with same seconds and nanos
+        """
+        if not isinstance(other, Timestamp):
+            return NotImplemented
```
parthban-db, line 312: ditto why notImplemeted.

databricks/sdk/common.py
hectorcast-db: Yes. In Python you cannot use a type before introduced. This is a workaround introduced at some point:

https://peps.python.org/pep-0484/#forward-references

databricks/sdk/common.py
hectorcast-db: https://docs.python.org/3/reference/datamodel.html#notimplemented
```
Numeric methods and rich comparison methods should return this value if they do not implement the operation for the operands provided. (The interpreter will then try the reflected operation, or some other fallback, depending on the operator.)
```

In short, if `a == b` (short for `a.__eq__(b)`) returns not implemented, python will try b == a (short for `b.__eq__(a)`). If both return not implemented, it will decide the result is false.

This way you can implement equality with objects from other libraries.

databricks/sdk/common.py
hectorcast-db: True. Fixing.

parthban-db (approved):

databricks/sdk/common.py
```diff
@@ -0,0 +1,348 @@
...
+"""
+
+from __future__ import annotations
+
+import logging
+import re
+from datetime import datetime, timedelta, timezone
+from decimal import Decimal
+
+_LOG = logging.getLogger("databricks.sdk")
```
renaudhartert-db: Do we need this?

databricks/sdk/common.py
```diff
@@ -0,0 +1,348 @@
+"""Common types for the Databricks SDK.
+
+This module provides common types used by different APIs.
+"""
```
renaudhartert-db: Could we actually move this in `databricks/sdk/core/common`? I'd like to reserve `databricks/sdk/common` for SDK-Mod.

databricks/sdk/common.py
hectorcast-db: Removed.

databricks/sdk/common.py
hectorcast-db: moved to `databricks/sdk/common_types/common.py` since moving it to `core` creates circular dependencies witth the `core.py` file.

hectorcast-db:
Not relevant.

PR #988, opened 2025-06-11
Don't use the global default logging instance when logging

mgyucht:

The Python SDK sometimes uses the global logging instance to log rather than a per-module logger instance, e.g.

https://github.com/databricks/databricks-sdk-py/blob/a3936022a8ba6fdd73d58b6e5647f226da7c5ae5/databricks/sdk/runtime/__init__.py#L116

Using logging.debug configures a root logger and is considered a bad practice when using the logging library in Python. This can lead to several issues including duplicated logs.

This PR makes this change for the few places where the default logger instance is used in the SDK.

## How is this tested?

**ALWAYS ANSWER THIS QUESTION:** Answer with "N/A" if tests are not applicable
to your PR (e.g. if the PR only modifies comments). Do not be afraid of 
answering "Not tested" if the PR has not been tested. Being clear about what 
has been done and not done provides important context to the reviewers.

### DIFF
--- a/databricks/sdk/_widgets/__init__.py
+++ b/databricks/sdk/_widgets/__init__.py
@@ -5,6 +5,7 @@ from abc import ABC, abstractmethod
 
 _LOG = logging.getLogger(__name__)
 
+
 class WidgetUtils(ABC):
 
     def get(self, name: str):
--- a/databricks/sdk/errors/customizer.py
+++ b/databricks/sdk/errors/customizer.py
@@ -5,6 +5,7 @@ import requests
 
 _LOG = logging.getLogger(__name__)
 
+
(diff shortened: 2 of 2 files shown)


databricks/sdk/runtime/__init__.py
```diff
@@ -3,7 +3,7 @@
 import logging
 from typing import Dict, Optional, Union, cast
 
-logger = logging.getLogger("databricks.sdk")
```
mgyucht, line 6: This existed before, but wasn't consistently used. I've renamed it to `_LOG` as it is not meant to be exposed publicly and it is consistent with other places within the SDK.

PR #991, opened 2025-06-14
Remove Unused Request Messages

parthban-db:
This PR removes Unused Request Messages. It is the result of resolving a bug in the code generator.

## How is this tested?

Existing CI.


parthban-db:
@renaudhartert-db @hectorcast-db, Do you think it is safe to merge this PR? Although these messages are never used in the SDKs, but still this removes a lot of unused messages.

PR #1006, opened 2025-07-17
Add headers to HttpRequestResponse in OpenAI client.

sunishsheth2009:
Adding headers to HttpRequestResponse when making http_request

Why?
Basically we need the response headers that is used in the MCP spec: https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management
Specially Mcp-Session-Id header.

## How is this tested?
```
%pip install git+https://github.com/sunishsheth2009/databricks-sdk-py@sunish-add-headers
import requests
import uuid
from databricks.sdk.service.serving import ExternalFunctionRequestHttpMethod

# Step 1: Initialize session
init_payload = {
    "jsonrpc": "2.0",
    "id": "init-1",
    "method": "initialize",
    "params": {}
}

init_response = WorkspaceClient().serving_endpoints.http_request(
  conn="github-u2m-managed-connection",
  method=ExternalFunctionRequestHttpMethod.POST,
  path="/mcp",
  json=init_payload,
)

session_id = init_response.headers["mcp-session-id"]
if not session_id:
    print("No session ID returned by server.")

print(f"✅ Got MCP Session ID: {session_id}")


Output:
✅ Got MCP Session ID: 53220122-015c-4e2b-9edb-47580cefffd0
```

### DIFF
--- a/databricks/sdk/mixins/open_ai_client.py
+++ b/databricks/sdk/mixins/open_ai_client.py
@@ -4,6 +4,7 @@ from typing import Dict, Optional
 from requests import Response
 
 from databricks.sdk.service.serving import (ExternalFunctionRequestHttpMethod,
+                                            HttpRequestResponse,
                                             ServingEndpointsAPI)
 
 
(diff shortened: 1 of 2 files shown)


databricks/sdk/mixins/open_ai_client.py
```diff
@@ -88,15 +89,30 @@ def http_request(
...
+                "connection_name": conn,
+                "method": method.value,
+                "path": path,
+                "headers": js.dumps(headers) if headers is not None else None,
+                "json": js.dumps(json) if json is not None else None,
+                "params": js.dumps(params) if params is not None else None,
+            },
+            headers={"Accept": "text/plain", "Content-Type": "application/json"},
+            raw=True,
+            response_headers=headers_to_capture,
```
renaudhartert-db, line 110: @hectorcast-db, can you think of something better given the current libraries? Not using the underlying API method is really not ideal but I can't think of a better alternative with the code that we have right now.

PR #1013, opened 2025-07-23
Update SDK with latest API changes.

renaudhartert-db:

This PR updates the SDK to include the latest API changes. 

## How is this tested?

Unit + Integration tests.


databricks/sdk/service/agentbricks.py
```diff
@@ -0,0 +1,342 @@
+# Code generated from OpenAPI specs by Databricks SDK Generator. DO NOT EDIT.
```
parthban-db: I see that agentbricks.py has been added, but I don't see that aibuilders.py has been deleted. AFAIK agentbricks is just a rename from aibuilders. You need to manually delete that file; it does not get deleted with the code generation.

parthban-db (approved):

PR #1025, opened 2025-08-12
Add support for passing additional kwargs when instantiating an OpenAI client for Databricks model serving

smurching:

Per customer request, adds support for passing additional optional kwargs to `databricks.sdk.WorkspaceClient().get_open_ai_client()`. This enables use cases like specifying a custom timeout and retry policy for latency-sensitive workloads, e.g:

```
>>> # Configure custom timeout and retries
>>> client = workspace_client.serving_endpoints.get_open_ai_client(
...     timeout=30.0, 
...     max_retries=5
... )
```

## How is this tested?

Updated unit tests

### DIFF
--- a/databricks/sdk/mixins/open_ai_client.py
+++ b/databricks/sdk/mixins/open_ai_client.py
@@ -3,11 +3,9 @@ from typing import Dict, Optional
 
 from requests import Response
 
-from databricks.sdk.service.serving import (
-    ExternalFunctionRequestHttpMethod,
-    HttpRequestResponse,
-    ServingEndpointsAPI,
-)
+from databricks.sdk.service.serving import (ExternalFunctionRequestHttpMethod,
+                                            HttpRequestResponse,
+                                            ServingEndpointsAPI)
 
 
 class ServingEndpointsExt(ServingEndpointsAPI):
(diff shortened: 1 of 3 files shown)


databricks/sdk/mixins/open_ai_client.py
```diff
@@ -31,19 +31,54 @@ def auth_flow(self, request: httpx.Request) -> httpx.Request:
...
-        )
+        # Default parameters that are required for Databricks integration
+        client_params = {
+            "base_url": self._api._cfg.host + "/serving-endpoints",
+            "api_key": "no-token",  # Passing in a placeholder to pass validations, this will not be used
+            "http_client": self._get_authorized_http_client(),
+        }
+
+        # Update with any additional parameters passed by the user
+        client_params.update(kwargs)
```
bbqiu, line 93: do we want to make it so you can't override the default client_params since it'll break compatibility w/ databricks model serving? we could also document what the defaults are

smurching: Yeah, seems safe to block overriding those & throw to start

hectorcast-db (approved):
LGTM

PR #1027, opened 2025-08-14
Add native support for Azure DevOps OIDC authentication

Divyansh-db:
Added native support for Azure DevOps OIDC authentication to allow authentication from Azure DevOps pipeline's environment.

## Testing
Tested on an Azure DevOps project. Used the SDK to authenticate with a Databrick's workspace using Azure DevOps OIDC

1) Created a demo python files that uses the SDK to create a Workspace Client: This will test if the OIDC authentication is working
Screenshot 2025-09-18 at 15 42 18

2) Created a pipeline to run the demo file. Set the necessary environment variables.
(System.AccessToken is slightly different, it needs to be exported through pipeline syntax: https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#systemaccesstoken) 
Screenshot 2025-09-18 at 15 43 34

3) The pipeline runs successfully indicating that authentication succeeded.
Screenshot 2025-09-18 at 15 43 59

### DIFF
--- a/databricks/sdk/credentials_provider.py
+++ b/databricks/sdk/credentials_provider.py
@@ -435,7 +435,9 @@ def github_oidc(cfg: "Config") -> Optional[CredentialsProvider]:
     Supported in GitHub Actions with OIDC service connections.
     """
     return _oidc_credentials_provider(
-        cfg=cfg, supplier_factory=lambda: oidc_token_supplier.GitHubOIDCTokenSupplier(), provider_name="GitHub OIDC"
+        cfg=cfg,
+        supplier_factory=lambda: oidc_token_supplier.GitHubOIDCTokenSupplier(),
+        provider_name="GitHub OIDC",
     )
 
 
--- a/databricks/sdk/oidc_token_supplier.py
+++ b/databricks/sdk/oidc_token_supplier.py
@@ -7,6 +7,7 @@ import requests
 logger = logging.getLogger("databricks.sdk")
 
 
+# TODO: Check the required environment variables while creating the instance rather than in the get_oidc_token method to allow early return.
 class GitHubOIDCTokenSupplier:
     """
     Supplies OIDC tokens from GitHub Actions.


databricks/sdk/credentials_provider.py
```diff
@@ -89,7 +89,6 @@ def inner(
         @functools.wraps(func)
         def wrapper(cfg: "Config") -> Optional[CredentialsProvider]:
             for attr in require:
-                getattr(cfg, attr)
```
Divyansh-db, line 92: Seemed like unnecessary Double check

databricks/sdk/credentials_provider.py
```diff
@@ -1003,18 +1053,19 @@ def model_serving_auth(cfg: "Config") -> Optional[CredentialsProvider]:
 
 
 class DefaultCredentials:
-    """Select the first applicable credential provider from the chain"""
```
Divyansh-db, line 1007: Strategy seems to be the correct word as we are trying different auth strategies until one suceeds

databricks/sdk/oidc_token_supplier.py
```diff
@@ -26,3 +26,57 @@ def get_oidc_token(self, audience: str) -> Optional[str]:
...
+    Constructs the OIDC token request URL using official Azure DevOps predefined variables.
+    See: https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables
+    """
+
+    def get_oidc_token(self, audience: str) -> Optional[str]:
+        # Note: Azure DevOps OIDC tokens have a fixed audience of "api://AzureADTokenExchange"
+        # The audience parameter is ignored but kept for interface compatibility with other OIDC suppliers
+
+        # Check for required Azure DevOps environment variables
+        access_token = os.environ.get("SYSTEM_ACCESSTOKEN")
```
hectorcast-db: Any reason why this is not loaded in the config and read from there?

databricks/sdk/credentials_provider.py
hectorcast-db: I remember discussing this a while ago and the decision was that provider was better.

databricks/sdk/oidc_token_supplier.py
Divyansh-db: it was done the same way in github OIDC that's why. But other strategies use config so we can also put these environment variables in config

databricks/sdk/credentials_provider.py
Divyansh-db: I came after working in go SDK. It was named strategy there, therefore made the change here. But if this was already discussed then I will revert back to provider.

databricks/sdk/credentials_provider.py
parthban-db: Are you sure? what does this function do?

databricks/sdk/credentials_provider.py
```diff
@@ -100,16 +99,37 @@ def wrapper(cfg: "Config") -> Optional[CredentialsProvider]:
...
     def inner(
         func: Callable[["Config"], OAuthCredentialsProvider],
     ) -> OauthCredentialsStrategy:
         @functools.wraps(func)
         def wrapper(cfg: "Config") -> Optional[OAuthCredentialsProvider]:
+            # Early environment detection - check before config validation
+            if env_vars and not all(os.environ.get(var) for var in env_vars):
+                # Provide specific error message for Azure DevOps OIDC SYSTEM_ACCESSTOKEN
+                if (
+                    name == "azdo-oidc"
```
parthban-db: IIRC we named azure-devops-oidc in golang, we should maintain the consistency of it.

databricks/sdk/credentials_provider.py
```diff
@@ -100,16 +99,37 @@ def wrapper(cfg: "Config") -> Optional[CredentialsProvider]:
...
+                    and "SYSTEM_ACCESSTOKEN" in env_vars
+                    and not os.environ.get("SYSTEM_ACCESSTOKEN")
```
parthban-db: This doesn't seems to be the correct place for this block. As this is a general oauth_credentials_strategy function and not for some specific strategy.

databricks/sdk/credentials_provider.py
```diff
@@ -408,6 +428,71 @@ def token() -> oauth.Token:
     return OAuthCredentialsProvider(refreshed_headers, token)
 
 
+@oauth_credentials_strategy(
+    "azdo-oidc",
```
parthban-db: ditto name of strategy.

databricks/sdk/credentials_provider.py
```diff
@@ -408,6 +428,71 @@ def token() -> oauth.Token:
...
+    ["host", "client_id"],
+    env_vars=[
```
parthban-db: Why using a mix of named and positional arguments? I would recommend this to be all positional argument.

databricks/sdk/oidc_token_supplier.py
```diff
@@ -26,3 +29,60 @@ def get_oidc_token(self, audience: str) -> Optional[str]:
...
+        collection_uri = os.environ.get("SYSTEM_TEAMFOUNDATIONCOLLECTIONURI")
+        project_id = os.environ.get("SYSTEM_TEAMPROJECTID")
+        plan_id = os.environ.get("SYSTEM_PLANID")
+        job_id = os.environ.get("SYSTEM_JOBID")
+        hub_name = os.environ.get("SYSTEM_HOSTTYPE")
+
+        # Check for required variables
+        if not all([access_token, collection_uri, project_id, plan_id, job_id, hub_name]):
+            # not in Azure DevOps pipeline
```
parthban-db: Let's have proper statements for comments. + ditto all.

databricks/sdk/oidc_token_supplier.py
```diff
@@ -26,3 +29,60 @@ def get_oidc_token(self, audience: str) -> Optional[str]:
...
+            logger.debug("Azure DevOps OIDC: not in Azure DevOps pipeline environment")
+            return None
+
+        try:
+            # Construct the OIDC token request URL
```
parthban-db:
```suggestion
            # Construct the OIDC token request URL.
```
Comment should be proper sentence ending with a period.
+ ditto all

databricks/sdk/credentials_provider.py
Divyansh-db: This function checks if the attribute exists in the config or not. This check is redundant, as it is done in the next line as well

databricks/sdk/credentials_provider.py
```diff
@@ -408,6 +412,64 @@ def token() -> oauth.Token:
...
+@oauth_credentials_strategy("azure-devops-oidc", ["host", "client_id"])
+def azure_devops_oidc(cfg: "Config") -> Optional[CredentialsProvider]:
+    """
+    Azure DevOps OIDC authentication uses a Token Supplier to get a JWT Token
+    and exchanges it for a Databricks Token.
+
+    Supported in Azure DevOps pipelines with OIDC service connections.
+    """
+    try:
+        supplier = oidc_token_supplier.AzureDevOpsOIDCTokenSupplier()
```
hectorcast-db: Why is this needed? `GitHubOIDCTokenSupplier` does raise an exception, and therefore there is no try/catch. Is this different for a reason?

databricks/sdk/credentials_provider.py
```diff
@@ -408,6 +412,64 @@ def token() -> oauth.Token:
...
+        return None
+
+    audience = cfg.token_audience
+    if audience is None and cfg.is_account_client:
+        audience = cfg.account_id
+    if audience is None and not cfg.is_account_client:
+        audience = cfg.oidc_endpoints.token_endpoint
+
+    # Try to get an idToken. If no supplier returns a token, we cannot use this authentication mode.
+    id_token = supplier.get_oidc_token(audience)
```
hectorcast-db: This function is almost the same as `github_oidc`. Can and should we reuse the code?

databricks/sdk/credentials_provider.py
```diff
@@ -1016,6 +1078,7 @@ def __init__(self) -> None:
             env_oidc,
             file_oidc,
             github_oidc,
+            azure_devops_oidc,
```
hectorcast-db: Can you explain why this is added in this particular spot? Order here matters and the wrong order can break customers.

databricks/sdk/oidc_token_supplier.py
```diff
@@ -26,3 +29,79 @@ def get_oidc_token(self, audience: str) -> Optional[str]:
             return None
 
         return response_json["value"]
+
```
hectorcast-db, line 33: Same: GitHubOIDC does not validate on create. Is there a reasons to change this?

Divyansh-db: This is done to ensure Early exit. This is similar to what is done in Go SDK. If the environment variables are not set then we are sure that we are not in Azure DevOps Environment so we should exit at the earliest and try other providers.

databricks/sdk/credentials_provider.py
Divyansh-db: I placed it near the other OIDC providers. Should the new provider be added to the end of the list always?

databricks/sdk/credentials_provider.py
Divyansh-db: Same. For early exit.

databricks/sdk/credentials_provider.py
Divyansh-db: The old code should not be changed, right? The common part is within the GithubOIDC function. I can refactor the code to put the common parts outside it and use them in AzureDevOps. What is the better option?

hectorcast-db: On a quick look, I think that the only difference if the supplier being used. 
We may be able to have
```
#TODO: better name
def _internal_oidc(cfg: "Config", supplier: ) -> Optional[CredentialsProvider]:
...

@oauth_credentials_strategy("azure-devops-oidc", ["host", "client_id"])
def azure_devops_oidc(cfg: "Config") -> Optional[CredentialsProvider]:
    return _internal_oidc(cfg, () -> oidc_token_supplier.AzureDevOpsOIDCTokenSupplier())
```

databricks/sdk/credentials_provider.py
```diff
@@ -408,6 +426,34 @@ def token() -> oauth.Token:
...
+@oauth_credentials_strategy("github-oidc", ["host", "client_id"])
+def github_oidc(cfg: "Config") -> Optional[CredentialsProvider]:
+    """
+    GitHub OIDC authentication uses a Token Supplier to get a JWT Token and exchanges
+    it for a Databricks Token.
+
+    Supported in GitHub Actions with OIDC service connections.
+    """
+    return _oidc_credentials_provider(
+        cfg=cfg, supplier_factory=lambda: oidc_token_supplier.GitHubOIDCTokenSupplier(), provider_name="GitHub OIDC"
```
renaudhartert-db: This will tell the formatter to have one argument per line (like you do below).

```suggestion
        cfg=cfg, supplier_factory=lambda: oidc_token_supplier.GitHubOIDCTokenSupplier(), provider_name="GitHub OIDC",
```

databricks/sdk/oidc_token_supplier.py
renaudhartert-db: Could we add a TODO on the GitHub OIDC implementation to make it clear that it is the one with tech debt?

renaudhartert-db (approved):

PR #1041, opened 2025-09-10
Add FieldMask class and helper functions

hectorcast-db:
Add FieldMask class and helper functions.

This class methods have the same signature than the google.protobuf.FieldMask
https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/internal/field_mask.py#L18
but serialize the FieldMask using snake_case, as expected by Databricks server.

Helper functions are required because serialization framework in templates only supports single line code.

Version Pin 
1. Why >=4.21.0:
Stability: Version 4.21.0 was a stable release that introduced important performance improvements and API stabilization
Python 3.7+ compatibility.

2. Why <7.0:
Major version boundary: Avoid future breaking changes.

## How is this tested?
Added unit tests.

NO_CHANGELOG=true


tests/test_fieldmask.py
```diff
@@ -0,0 +1,167 @@
+import pytest
+
+from databricks.sdk.common.types.fieldmask import FieldMask
+
+
+def test_to_json_string_basic():
```
Divyansh-db: I think we should use table-driven tests.
(Example: https://github.com/databricks/databricks-sdk-go/pull/1292/files?w=1#diff-f9e861be2e9f308e900ba0c5eae590ac087cac190dd1a19b27d06610c99bd2d4R84)

hectorcast-db: Updated

Divyansh-db (approved):
LGTM, Can you also get an approval from Parth or Omer

databricks/sdk/common/types/fieldmask.py
```diff
@@ -0,0 +1,15 @@
+class FieldMask(object):
+    """Class for FieldMask message type."""
+
+    def ToJsonString(self):
```
parthban-db: Can we add a return type to this? 
```suggestion
    def ToJsonString(self) -> string:
```

databricks/sdk/common/types/fieldmask.py
```diff
@@ -0,0 +1,15 @@
...
+        """Converts FieldMask to string."""
+        return ",".join(self.paths)
+
+    def FromJsonString(self, value):
```
parthban-db: ditto 
```suggestion
    def FromJsonString(self, value: str) -> Array[str]:
```

Maybe my suggested types are a bit off.

databricks/sdk/service/_internal.py
```diff
@@ -1,6 +1,11 @@
 import datetime
 import urllib.parse
-from typing import Callable, Dict, Generic, Optional, Type, TypeVar
+from typing import Callable, Dict, Generic, List, Optional, Type, TypeVar
+
+from google.protobuf.duration_pb2 import Duration
+from google.protobuf.timestamp_pb2 import Timestamp
```
parthban-db, line 6: Do we need to add this dependency somewhere? Or are they already added? The thing I want to know is whether we are pinning the version of these dependencies?

databricks/sdk/service/_internal.py
```diff
@@ -46,6 +51,63 @@ def _escape_multi_segment_path_parameter(param: str) -> str:
     return urllib.parse.quote(param)
 
 
+def _timestamp(d: Dict[str, any], field: str) -> Optional[Timestamp]:
```
parthban-db, line 54: Can we add a doc comment to this function?

tests/test_internal.py
```diff
@@ -52,3 +58,141 @@ def test_escape_multi_segment_path_parameter():
...
+
+    if expected_result is None:
+        assert result is None
+    else:
+        assert isinstance(result, Timestamp)
+        assert result.ToJsonString() == expected_json
+
+
+@pytest.mark.parametrize(
+    "input_dict,field_name,expected_length,expected_json_list,description",
```
parthban-db: Do we need to have two parameters, expected_length and expected_json_list, as the first can be derived from the second?

tests/test_internal.py
```diff
@@ -52,3 +58,141 @@ def test_escape_multi_segment_path_parameter():
...
+
+    if expected_result is None:
+        assert result is None
+    else:
+        assert isinstance(result, FieldMask)
+        assert result.ToJsonString() == expected_json
+
+
+@pytest.mark.parametrize(
+    "input_dict,field_name,expected_length,expected_json_list,description",
```
parthban-db: ditto

tests/test_internal.py
```diff
@@ -52,3 +58,141 @@ def test_escape_multi_segment_path_parameter():
...
+)
+def test_timestamp(input_dict, field_name, expected_result, expected_json, description):
+    """Test _timestamp function with various input scenarios."""
+    result = _timestamp(input_dict, field_name)
+
+    if expected_result is None:
+        assert result is None
+    else:
+        assert isinstance(result, Timestamp)
+        assert result.ToJsonString() == expected_json
```
parthban-db: Can we simply check the equality of the Timestamp instead of converting it to JSON?

tests/test_internal.py
```diff
@@ -52,3 +58,141 @@ def test_escape_multi_segment_path_parameter():
...
+    """Test _repeated_timestamp function with various input scenarios."""
+    result = _repeated_timestamp(input_dict, field_name)
+
+    if expected_length is None:
+        assert result is None
+    else:
+        assert len(result) == expected_length
+        assert all(isinstance(ts, Timestamp) for ts in result)
+        for i, expected_json in enumerate(expected_json_list):
+            assert result[i].ToJsonString() == expected_json
```
parthban-db: ditto all

tests/test_internal.py
```diff
@@ -52,3 +58,141 @@ def test_escape_multi_segment_path_parameter():
...
+
+    if expected_result is None:
+        assert result is None
+    else:
+        assert isinstance(result, Duration)
+        assert result.ToJsonString() == expected_json
+
+
+@pytest.mark.parametrize(
+    "input_dict,field_name,expected_length,expected_json_list,description",
```
parthban-db: ditto expected_length can be derived from the expected_json_list.

databricks/sdk/common/types/fieldmask.py
hectorcast-db: I will add the hints.
Note that this method modifies the element itself and returns None.
As a reference, this follows the same pattern and signature than the original class
https://pigweed.googlesource.com/third_party/github/protocolbuffers/protobuf/+/HEAD/python/google/protobuf/internal/field_mask.py
(added as a comment)

tests/test_internal.py
```diff
@@ -52,3 +58,156 @@ def test_escape_multi_segment_path_parameter():
...
+)
+def test_timestamp(input_dict, field_name, expected_result, expected_timestamp, description):
+    """Test _timestamp function with various input scenarios."""
+    result = _timestamp(input_dict, field_name)
+
+    if expected_result is None:
+        assert result is None
+    else:
+        assert isinstance(result, Timestamp)
+        assert result == expected_timestamp
```
parthban-db: I think we don't need to have this expected_result now, as the expected_timestamp will fulfill the purpose. As it can be None on its own. + ditto all

parthban-db (approved):
LGTM modulo nit.

Issue #1043, opened 2025-09-15
[FEATURE] Add `served-model` support to serving endpoint query api


aidint:
**Problem Statement**
Both `serving_endpoints.query` and `serving_endpoints_data_plane.query` on `WorkspaceClient` do not support querying individual models behind endpoints as explained [here](https://docs.databricks.com/aws/en/machine-learning/model-serving/serve-multiple-models-to-serving-endpoint#query-individual-models-behind-an-endpoint).

**Proposed Solution**
Add "served_model_name" parameter to the query method on both serving_endpoint classes.
If served_model_name is provided, then replace the request url with: `url.replace("/invocations", f"/served-models/{served_model_name}")`

PR #1044, opened 2025-09-15
Add served-model support to serving endpoint query

aidint:

This PR adds support for querying an individual model in the serving endpoints as documented [here](https://docs.databricks.com/aws/en/machine-learning/model-serving/serve-multiple-models-to-serving-endpoint#query-individual-models-behind-an-endpoint).

- **WHAT** changes are being made in the PR? 
It adds served-models to the url used for the query requests if `served_model_name` is provided.  
- **WHY** are these changes needed?
While querying an individual model behind a serving endpoint is supported by rest api, it is not supported by the sdk. This limits users flexibility in serving multiple models in one endpoint. I raised this problem in [this issue](https://github.com/databricks/databricks-sdk-py/issues/1043).

## How is this tested?
I tested this with an endpoint in a private databricks workspace.

I couldn't find any test case for the query methods, therefore the answer is N/A.

### DIFF
--- a/databricks/sdk/service/serving.py
+++ b/databricks/sdk/service/serving.py
@@ -4505,6 +4505,7 @@ class ServingEndpointsAPI:
         stream: Optional[bool] = None,
         temperature: Optional[float] = None,
         usage_context: Optional[Dict[str, str]] = None,
+        served_model_name: Optional[str] = None,
     ) -> QueryEndpointResponse:
         """Query a serving endpoint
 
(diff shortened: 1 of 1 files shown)


aidint:
@hectorcast-db can you please help me with this PR?

aidint:
@rauchy  can you please review this pull request?

hectorcast-db:
Hi @aidint 
Thanks for reaching out and the PR. 
Unfortunately, we cannot merge this PR. The code for services is auto generated, so any manual change will be automatically reverted. 
I will forward this to the team so they can update their configuration upstream.

PR #1056, opened 2025-09-30
Fix: ensure dbutils.fs.ls lists directories with trailing '/'

shindeakshay0211:
Fixes local directory listing to include folders as well as files. Directory paths now end with '/' and are correctly marked as directories.


shindeakshay0211:
fix: include directories in local file listing and update changelog

PR #1060, opened 2025-10-03
Updated release workflow

Divyansh-db:

Updated the Release workflow. 

- Now it creates an actual release instead of draft
- It also automatically pulls the Next changelog info from the tag and uses them in the Release notes.

Deleted some unused workflows:

- release-test.yml - run once in feb, 2025

## How is this tested?
N/A

NO_CHANGELOG=true

### DIFF
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -1,0 +1,37 @@
+name: Release (TestPyPI)
+
+on:
+  workflow_dispatch:
+
+jobs:
+  publish:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-test.yml
```diff
@@ -1,37 +0,0 @@
-name: Release (TestPyPI)
-
```
Divyansh-db, line 2: @mgyucht Is this workflow still relevant/required?

PR #1082, opened 2025-10-22
[Bug Fix] Return error message when body has no length when logging e…

tanmay-db:
# WIP

When logging error, we check for empty body by using len(), however this doesn't work for example with `_io.BytesIO`
`TypeError: object of type '_io.BytesIO' has no len()`

## How is this tested?
- Will add if this is the preferred approach

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Bug Fixes
 - Improving the error message that is shown when the unsupported `dbutils.credentials.getServiceCredentialsProvider` method is used. This method can only be used inside of a notebook.
-- Return error message when body doesn't support `len()` while logging error.
+- Check for string body type while logging error message.
 
 ### Documentation
 


databricks/sdk/logger/round_trip_logger.py
```diff
@@ -110,8 +110,12 @@ def _recursive_marshal(self, v: Any, budget: int) -> Any:
             return v
 
     def _redacted_dump(self, prefix: str, body: str) -> str:
-        if len(body) == 0:
-            return ""
+        try:
```
serena-ruan: I think we should check `isinstance(body, str)` instead of try catch

serena-ruan: Also could we try `str(body)`?

tanmay-db: right, I think isinstance would be better.

databricks/sdk/logger/round_trip_logger.py
```diff
@@ -110,6 +110,8 @@ def _recursive_marshal(self, v: Any, budget: int) -> Any:
             return v
 
     def _redacted_dump(self, prefix: str, body: str) -> str:
+        if not isinstance(body, str):
+            return "unsupported body type"
```
renaudhartert-db, line 114: The method signature says that this should be a string. Either adjust the type to cover all potential cases or make sure that the method is only called with string inputs.

NEXT_CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### Bug Fixes
 - Improving the error message that is shown when the unsupported `dbutils.credentials.getServiceCredentialsProvider` method is used. This method can only be used inside of a notebook.
+- Check for string body type while logging error message.
```
renaudhartert-db, line 9: I think we could have a message that explains a little more what the change is about. For example, as a reader, I'm wondering "what does checking for string body type" actually does? I believe something like the following would more useful:

> Fix a `TypeError` error when logging non-string body types in `RoundTrip`.

Also, could you update the change log above so that it phrased at the imperative tense? That is "Improve the error message ..."

tanmay-db: Hi @renaudhartert-db, thanks for the review. 

The PR is WIP and I just added a short line so it doesn't fail check for next changelog test, was planning to elaborate it and add test before sending it to review.

tanmay-db:
closing this since @serena-ruan has a fix for this issue.

PR #1088, opened 2025-10-27
Fix FilesExt upload fails when content size is zero

yuanjieding-db:

Provide the readers and reviewers with the information they need to understand
this PR in a comprehensive manner. 

- **WHAT** 
  - Fix the `FilesExt` issue where it throws exception when the content size is 0, by checking if the size of the content is below a certain threshold. If it is below the threshold, upload using single-shot instead of multipart upload.
- **WHY** 
  - `upload` interface needs to support upload data with zero size.

## How is this tested?

Unit test cases are updated to test the behavior change.

### DIFF
--- a/tests/test_files.py
+++ b/tests/test_files.py
@@ -1806,7 +1806,7 @@ class MultipartUploadTestCase(UploadTestCase):
         ),
         MultipartUploadTestCase(
             "Multipart upload successful: empty file or empty seekable stream",
-            content_size=0,  # less than part size
+            content_size=0,  # content with zero length
             multipart_upload_min_stream_size=100 * 1024 * 1024,  # all files smaller than 100M goes to single-shot
             expected_single_shot_upload=True,
             expected_multipart_upload_aborted=None,


yuanjieding-db:
jenkins trigger all

databricks/sdk/mixins/files.py
```diff
@@ -1206,7 +1208,7 @@ def upload_from(
             use_parallel=use_parallel,
             parallelism=parallelism,
         )
-        if ctx.use_parallel:
+        if ctx.use_parallel and ctx.content_length >= self._config.files_ext_multipart_upload_min_stream_size:
```
parthban-db, line 1211: Can content_length be none here?

tests/test_files.py
```diff
@@ -1738,10 +1746,22 @@ def to_string(test_case: "MultipartUploadTestCase") -> str:
...
+            source_type=[
+                UploadSourceType.FILE,
+                UploadSourceType.SEEKABLE_STREAM,
+            ],  # non-seekable streams always use multipart upload
             expected_part_size=1024 * 1024,  # chunk size is used
+            expected_single_shot_upload=True,
+        ),
+        MultipartUploadTestCase(
+            "Multipart upload successful: empty file or empty seekable stream",
+            content_size=0,  # less than part size
```
parthban-db:
```suggestion
            content_size=0,  # content with zero length
```

databricks/sdk/mixins/files.py
yuanjieding-db: No, in the `upload_from` function, `content_length` will always be set using `os.path.getsize(source_path)`, which only returns a number.

tests/test_files.py
yuanjieding-db: Updated.

databricks/sdk/mixins/files.py
parthban-db: Maybe, not in this PR. Let's avoid using the same uploadContext in both methods, as they may differ for different functions.
FWIW, I am planning to add a static type checker in the Python SDK, which will block such scenarios.

PR #1094, opened 2025-10-29
Support Scopes and Authorization Details in Workspace Client

aravind-segu:
This PR introduces two new high level parameters to WorkspaceClient -> Scopes and Authorization Details
- These parameters will be propogated to OAuth Authentication Methods
- The PR Updates current OAuth Authentication Methods to take in scopes and authorization details when retrieving tokens
- Currently the databricks-cli OAuth Credential Strategy does not support scopes and authorization details as this work is still in progress
- Introduces a new runtime-oauth credential strategy
     -  If scopes are specified in notebook environments, the pat token is exchanged for oauth token
- If no scopes are mentioned then traditional notebook authentication is performed
- In Default Credentials runtime-oauth now takes precedence so that it is checked first before falliing back to pat token authentication
- Token exchange API does not support `all-apis` due to security constraints, so runtime-native-auth is still maintained

## How is this tested?
- Added Unit tests for Notebook OAuth
- Need some pointers or repro docs to E2E test all OAuth providers

#### Scopes respected correctly
Here since the token has serving.serving-endpoints, clusters.list returns an error
image

When added `compute.clusters` both methods run successfully
image

### Fallback to Notebook PAT correctly
image

### Token cache is used correctly and the same token is returned if its not expired
image

### DIFF
--- a/databricks/sdk/config.py
+++ b/databricks/sdk/config.py
@@ -113,6 +113,7 @@ class Config:
     disable_experimental_files_api_client: bool = ConfigAttribute(
         env="DATABRICKS_DISABLE_EXPERIMENTAL_FILES_API_CLIENT"
     )
+    # TODO: Expose these via environment variables too.
     scopes: str = ConfigAttribute()
     authorization_details: str = ConfigAttribute()
 
--- a/databricks/sdk/oauth.py
+++ b/databricks/sdk/oauth.py
@@ -760,6 +760,22 @@ class ClientCredentials(Refreshable):
 
 @dataclass
 class PATOAuthTokenExchange(Refreshable):
+    """Performs OAuth token exchange using a Personal Access Token (PAT) as the subject token.
+
+    This class implements the OAuth 2.0 Token Exchange flow (RFC 8693) to exchange a Databricks
+    Internal PAT Token for an access token with specific scopes and authorization details.
+
+    Args:
+        get_original_token: A callable that returns the PAT to be exchanged. This is a callable
+            rather than a string value to ensure that a fresh Internal PAT Token is retrieved
+            at the time of refresh.
+        host: The Databricks workspace URL (e.g., "https://my-workspace.cloud.databricks.com").
+        scopes: Space-delimited string of OAuth scopes to request (e.g., "all-apis offline_access").
+        authorization_details: Optional JSON string containing authorization details as defined in
+            AuthorizationDetail class above.
+        disable_async: Whether to disable asynchronous token refresh. Defaults to True.
+    """
+
     get_original_token: Callable[[], Optional[str]]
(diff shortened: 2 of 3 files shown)


databricks/sdk/oauth.py
```diff
@@ -706,18 +730,21 @@ class ClientCredentials(Refreshable):
     client_secret: str
     token_url: str
     endpoint_params: dict = None
-    scopes: List[str] = None
+    scopes: str = None
```
renaudhartert-db, line 733: This looks like a regression to me. Before, it was clear that `scopes` was a list of individual scopes. Now, I don't know what `scopes` is. If it is a list of concatenated scopes, what is the separator? We should not expect users of the class to have to read the code to understand the meaning of a field.

databricks/sdk/oauth.py
```diff
@@ -706,18 +730,21 @@ class ClientCredentials(Refreshable):
...
     use_params: bool = False
     use_header: bool = False
     disable_async: bool = True
+    authorization_details: str = None
```
renaudhartert-db, line 737: Similar comment here.

databricks/sdk/config.py
```diff
@@ -113,6 +113,8 @@ class Config:
     disable_experimental_files_api_client: bool = ConfigAttribute(
         env="DATABRICKS_DISABLE_EXPERIMENTAL_FILES_API_CLIENT"
     )
+    scopes: str = ConfigAttribute()
+    authorization_details: str = ConfigAttribute()
```
renaudhartert-db, line 118:
```suggestion
    # TODO: Expose these via environment variables too.
    scopes: str = ConfigAttribute()
    authorization_details: str = ConfigAttribute()
```

renaudhartert-db, line 118: Could we make these a list and a dict instead?

renaudhartert-db: Discussed offline, we cannot easily use higher-level types in the config without a significant refactoring of the ConfigAttribute logic. It does not seem worth investing in such refactoring at this point in time.

databricks/sdk/oauth.py
```diff
@@ -731,6 +758,51 @@ def refresh(self) -> Token:
         )
 
 
+@dataclass
+class PATOAuthTokenExchange(Refreshable):
```
renaudhartert-db, line 762: Please document the class and the parameters. In particular, make it clear that `scopes` and `authorization_details` are `str` and explain why `get_original_token` is a Callable and not a token.

databricks/sdk/oauth.py
renaudhartert-db: Closing as per above response.

PR #1107, opened 2025-11-05
Fix `WorkspaceClient` example in the main `README.md` file.

renaudhartert-db:

Fix example in the main `README.md` file. 

## How is this tested?

N/A

NO_CHANGELOG=true


README.md
```diff
@@ -239,7 +239,8 @@ For example, to turn on debug HTTP headers:
 
 ```python
-w = WorkspaceClient(debug_headers=True)
+from databricks.sdk.core import Config
+w = WorkspaceClient(config=Config(debug_headers=True))
```
Divyansh-db: Why is this required? The WorkspaceClient takes debug_headers as an input

renaudhartert-db: True, but it does not work with all the properties listed above (e.g. `retry_timeout_seconds`). Doing it with a Config always work. I've updated the example to use a parameter that only exist at the `Config` level.

PR #1115, opened 2025-11-12
Add fall back to all exceptions encountered when presigned URL is not available for upload and download in FilesExt

yuanjieding-db:

**WHAT**
- Add fall back to all exceptions encountered when presigned URL is not available for upload and download in `FilesExt`

**WHY**
- This is to eliminate the possibility of any potential regression introduced by upload and download through presigned URL.

## How is this tested?

Unit tests are updated to reflect the change.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Release v0.74.0
 
 ### New Features and Improvements
+* Add new auth type (`runtime-oauth`) for notebooks: Introduce a new authentication mechanism that allows notebooks to authenticate using OAuth tokens
 
 ### Security
 
--- a/databricks/sdk/__init__.py
+++ b/databricks/sdk/__init__.py
@@ -1,7 +1,8 @@
 # Code generated from OpenAPI specs by Databricks SDK Generator. DO NOT EDIT.
 
+import json
 import logging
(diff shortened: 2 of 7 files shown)


databricks/sdk/mixins/files.py
```diff
@@ -1466,7 +1499,14 @@ def _parallel_multipart_upload_from_stream(
...
             etag = self._do_upload_one_part(
-                ctx, cloud_provider_session, 1, 0, len(pre_read_buffer), session_token, BytesIO(pre_read_buffer)
+                ctx,
+                cloud_provider_session,
+                1,
+                0,
+                len(pre_read_buffer),
+                session_token,
+                BytesIO(pre_read_buffer),
+                is_first_part=True,
```
parthban-db, line 1509: Do we need to pass this information? Every error returned by this will fall back to FilesAPI, right?

yuanjieding-db: We do need this information for two reasons:
- Whenever it is not possible to use Presigned URL, the SDK should fail at the first part. Otherwise client should retry.
- For uploading of unseekable streams, if the function fails at any part other than the first part, the previous parts would been dropped from memory and thus not possible to perform a fallback. In that case the SDK has no option but throw an exception, and the client should perform a retry.

PR #1166, opened 2025-12-23
Config changes for custom scopes support

tejaskochar-db:
## Summary
- change type of `scopes` from `str` to `list`.
- add option for disabling refresh token to be used in U2M.
- sort scopes when config is initialized.
- add getters for scopes that add default scope "all-apis" if `scopes` is empty.

## How is this tested?
 - Unit tests for loading scopes from a test .databrickscfg config file
 - Unit tests ensuring `disable_oauth_refresh_token` is set when environment variable `DATABRICKS_DISABLE_OAUTH_REFRESH_TOKEN` is set.
---
NO_CHANGELOG=true

### DIFF
--- a/databricks/sdk/config.py
+++ b/databricks/sdk/config.py
@@ -75,7 +75,7 @@ def _parse_scopes(value):
         result = sorted(set(s for s in value if s))
         return result if result else None
     if isinstance(value, str):
-        parsed: list = sorted(set(s for s in re.split(r'[, ]+', value) if s))
+        parsed: list = sorted(set(s for s in re.split(r"[, ]+", value) if s))
         return parsed if parsed else None
     return None
 


tests/test_notebook_oauth.py
```diff
@@ -78,10 +78,10 @@ def credentials_provider() -> Dict[str, str]:
 @pytest.mark.parametrize(
     "scopes,auth_details",
     [
-        ("sql offline_access", None),
-        ("sql offline_access", '{"type": "databricks_resource"}'),
+        ("sql, offline_access", None),
+        ("sql, offline_access", '{"type": "databricks_resource"}'),
         ("sql", None),
-        ("sql offline_access all-apis", None),
+        ("sql, offline_access, all-apis", None),
```
tejaskochar-db: This is probably a breaking change, to avoid making it we will have to also allow for whitespace as a delimiter as runtime oauth already uses that format.

databricks/sdk/__init__.py
```diff
@@ -253,7 +253,7 @@ def __init__(
                 product=product,
                 product_version=product_version,
                 token_audience=token_audience,
-                scopes=" ".join(scopes) if scopes else None,
+                scopes=scopes,
```
hectorcast-db, line 256: This is a generated file. Have you changed the template? Ping me the PR for the template update.

databricks/sdk/config.py
```diff
@@ -133,10 +148,14 @@ class Config:
...
     )
-    # TODO: Expose these via environment variables too.
-    scopes: str = ConfigAttribute()
+
+    scopes: list = ConfigAttribute(transform=_parse_scopes)
     authorization_details: str = ConfigAttribute()
 
+    # disable_oauth_refresh_token controls whether a refresh token should be requested
+    # during the U2M authentication flow (default to false).
+    disable_oauth_refresh_token: bool = ConfigAttribute(env="DATABRICKS_DISABLE_OAUTH_REFRESH_TOKEN")
```
hectorcast-db, line 158: Are we using this?

tests/test_config.py
```diff
@@ -453,3 +453,42 @@ def test_no_org_id_header_on_regular_workspace(requests_mock):
...
+    mocker.patch("databricks.sdk.config.Config.init_auth")
+    monkeypatch.setenv("DATABRICKS_DISABLE_OAUTH_REFRESH_TOKEN", "true")
+    config = Config(host="https://test.databricks.com")
+    assert config.disable_oauth_refresh_token is True
+
+
+def test_disable_oauth_refresh_token_defaults_to_false(mocker):
+    mocker.patch("databricks.sdk.config.Config.init_auth")
+    config = Config(host="https://test.databricks.com")
+    assert config.disable_oauth_refresh_token is None  # ConfigAttribute returns None when not set
```
hectorcast-db: Why not checking 
```
assert not config.disable_oauth_refresh_token 
```
Is there a case which we want to differentiate between False and Not set?

tests/test_notebook_oauth.py
hectorcast-db: We should add that in this PR then. I see we keep the tests as they are and add new tests for comma.

hectorcast-db: Side note: space delimited is the standard according to [rfc6749](https://datatracker.ietf.org/doc/html/rfc6749). We may want to consider supporting them in other SDKs too.

tejaskochar-db: 👍

databricks/sdk/config.py
tejaskochar-db: In the next PR, yes.

tests/test_notebook_oauth.py
tejaskochar-db: We had discussed this and had chosen to only support commas as delimiter.

There are three stages for scopes right now:
1. The input that users provide. In INI files, the convention is to use commas to delimit lists. This is why we chose comma only. The python SDK has an extra runtime OAuth auth type, which was added very recently and takes scopes as a whitespace separated str. 
2. Internally, scopes are represented as lists.
3. When making requests, scopes are formatted as a whitespace separated string according to the RFC.

I will add whitespace as a delimiter here, but document that it is for backwards compatibility with runtime OAuth. In the meantime, we can re-open the discussion on whether we should also use whitespace as a delimiter in the other SDKs for consistency, or perhaps change all the SDKs to only use whitespace (since comma is only being introduced now, we can still change it) or use whitespace only in python.

PR #1211, opened 2026-01-20
update FilesExt retry logic

yuanjieding-db:

**WHAT**
- Extending retry function with a new parameter `max_attempt` to allow client to retry and fail after certain amount
- Remove 500 from FilesExt retry status code
- Add new config to set the retry attempts for FilesExt
- Update the retry logic of FilesExt to fail after certain attempts.

**WHY**
- 500 errors shouldn't be retried
- The FilesExt should always prioritize fallback over retry to avoid regression

## How is this tested?

Unit tests were updated to reflect the change.


databricks/sdk/config.py
```diff
@@ -233,6 +233,11 @@ class Config:
     # Cap on the number of custom retries during parallel downloads.
     files_ext_parallel_download_max_retries = 3
 
+    # Maximum number of retry attempts for FilesExt cloud API operations.
+    # This works in conjunction with retry_timeout_seconds - whichever limit
+    # is hit first will stop the retry loop.
+    files_ext_cloud_api_max_retries: int = 3
```
parthban-db: Maybe we can use this as a temporary parameter, since our end goal is not to fallback. 
```suggestion
    # Maximum number of retry attempts for FilesExt cloud API operations.
    # This works in conjunction with retry_timeout_seconds - whichever limit
    # is hit first will stop the retry loop.
    experimental_files_ext_cloud_api_max_retries: int = 3
```

databricks/sdk/retries.py
```diff
@@ -64,6 +65,10 @@ def wrapper(*args, **kwargs):
 
                     clock.sleep(sleep + random())
                     attempt += 1
+
+            # Determine which limit was hit
+            if max_attempts is not None and attempt > max_attempts:
+                raise TimeoutError(f"Exceeded max retry attempts ({max_attempts})") from last_err
```
parthban-db: Do we have a better error to represent this error? TimeoutError feels a bit odd for this case, as the function is not actually timed out.

tests/test_retries.py
```diff
@@ -293,3 +293,125 @@ def fn() -> Tuple[Any, Optional[RetryError]]:
...
+    @retried(
+        is_retryable=lambda _: "always retry",
+        timeout=timedelta(seconds=2),  # Short timeout
+        max_attempts=100,  # High max attempts
+        clock=clock,
+    )
+    def failing_function():
+        nonlocal call_count
+        call_count += 1
+        raise ValueError("test error")
```
parthban-db: Can we sleep on this failing function for a sec? I am a bit worried that this test will become flaky because it is possible to retry 100 times in 2 seconds.

tests/test_retries.py
```diff
@@ -293,3 +293,125 @@ def fn() -> Tuple[Any, Optional[RetryError]]:
...
+            raise ValueError("not yet")
+        return "success"
+
+    result = eventually_succeeds()
+    assert result == "success"
+    assert call_count == 3
+
+
+def test_max_attempts_none_preserves_backward_compatibility():
+    """Test that max_attempts=None only uses timeout (backward compatibility)."""
```
parthban-db:
```suggestion
def test_max_attempts_none():
    """Test that max_attempts=None only uses timeout."""
```

I think we don't need to mention that this test is for backward compatibility because, almost always, a unit test is for regression catching.

tests/test_retries.py
```diff
@@ -293,3 +293,125 @@ def fn() -> Tuple[Any, Optional[RetryError]]:
             assert clock.time() >= min_time
         elif scenario in ("halt", "unexpected"):
             assert call_count == attempts
+
+
+def test_max_attempts_respected():
```
parthban-db: Can we make these a table test to simplify the tests?

tests/test_retries.py
```diff
@@ -293,3 +293,125 @@ def fn() -> Tuple[Any, Optional[RetryError]]:
...
+    )
+    def failing_function():
+        nonlocal call_count
+        call_count += 1
+        raise ValueError("test error")
+
+    with pytest.raises(TimeoutError) as exc_info:
+        failing_function()
+
+    # Should have attempted 3 times (initial + 2 retries)
```
parthban-db:
```suggestion
    # Should have attempted 3 times (initial + 2 retries).
```

Period after a sentence, ditto all.

databricks/sdk/retries.py
yuanjieding-db: Ideally we should use a custom type `RetryError`, with `TimeoutError` and `MaxRetryExceededError` as it's derived types, so that the user can catch the `RetryError` if they don't care why the retry exhausted, while keeping the information.
However since we have been using built-in `TimeoutError` and users may already be depending on this behavior, it is risky to change it to a different Error.
If we were to introduce a new type of error for max retry exceeded scenario, it would be more difficult for the upper layer to handle the retry error: it needs to catch both Errors manually.

I don't see a better solution here, unless we can rewrite the `retry` logic completely, or make `FilesExt` using a different retry library.

databricks/sdk/config.py
yuanjieding-db: Updated

tests/test_retries.py
yuanjieding-db: Updated

tests/test_retries.py
yuanjieding-db: It wouldn't because of the backoff logic, right?

tests/test_retries.py
yuanjieding-db: Updated.

databricks/sdk/retries.py
parthban-db: So, you finally changed this to RuntimeError, right? I think it is fine to throw that, as I don't want any other function to depend on that error.

parthban-db (approved):

PR #1229, opened 2026-01-28
Use CLOUD_PROVIDER instead of host to determine which tests to run

hectorcast-db:
Use CLOUD_PROVIDER instead of host to determine which tests to run. 

This decouples test run from host parsing and allows testing Cloud specific features on cloud agnostic hosts.

## How is this tested?
Check test run and validate that the test runs only when expected
Screenshot 2026-01-28 at 10 54 49
Screenshot 2026-01-28 at 10 55 34

NO_CHANGELOG=true

### DIFF
--- a/tests/integration/test_deployment.py
+++ b/tests/integration/test_deployment.py
@@ -2,9 +2,13 @@ import logging
 
 import pytest
 
+from databricks.sdk.environments import Cloud
+
+from .conftest import _is_cloud
+
 
 def test_workspaces(a):
-    if a.config.is_azure:
+    if _is_cloud(Cloud.AZURE):
         pytest.skip("not available on Azure")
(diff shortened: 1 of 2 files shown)


tests/integration/test_deployment.py
```diff
@@ -2,9 +2,13 @@
...
 import pytest
 
+from databricks.sdk.environments import Cloud
+
+from .conftest import _is_cloud
+
 
 def test_workspaces(a):
-    if a.config.is_azure:
+    if _is_cloud(Cloud.AZURE):
```
tanmay-db, line 11: We only check for azure? no aws or gcp?

hectorcast-db: We actually have a single test in Python which is cloud specific.

PR #1232, opened 2026-01-30
 Enable support for a single Profile to use both Account and Workspace endpoints

hectorcast-db:
Enable support for a single Profile to use both Account and Workspace endpoints.

This is done by moving the header injection logic to the code generation, which is aware of which endpoints are workspace specific.

## How is this tested?
Updated test.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -84,10 +84,8 @@ def ucacct(env_or_skip) -> AccountClient:
 
 @pytest.fixture(scope="session")
 def unified_config(env_or_skip) -> Config:
-    _load_debug_env_if_runs_from_ide("workspace")
+    _load_debug_env_if_runs_from_ide("account")
     env_or_skip("CLOUD_ENV")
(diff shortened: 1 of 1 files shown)


tests/integration/test_unified_profile.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def test_workspace_operations(unified_config):
+    client = WorkspaceClient(config=unified_config)
+    user = client.current_user.me()
+    assert user is not None
+
+
+def test_account_operations(unified_config):
+    client = AccountClient(config=unified_config)
```
tanmay-db, line 11: Won't this test be skipped because the unified config above is defined only for workspace?

````
    _load_debug_env_if_runs_from_ide("workspace")
    env_or_skip("CLOUD_ENV")
    if "DATABRICKS_ACCOUNT_ID" in os.environ:
        pytest.skip("Skipping workspace test on account level")
````

tanmay-db (approved):

PR #1240, opened 2026-02-02
Enable Legacy Profiles/Config in Unified Mode

hectorcast-db:
This PR adds support for Legacy Profiles/Config in Unified Mode. 

This allows us to test the detection/fallback logic and the changes happening to the legacy workspace logic.

## How is this tested?
Run integration tests twice: one with unified flag and one without it.

### DIFF
--- a/tests/test_open_ai_mixin.py
+++ b/tests/test_open_ai_mixin.py
@@ -92,7 +92,13 @@ def test_langchain_open_ai_client(monkeypatch):
     assert client.model_name == "databricks-meta-llama-3-1-70b-instruct"
 
 
-def test_http_request(w, requests_mock):
+def test_http_request(monkeypatch, requests_mock):
+    from databricks.sdk import WorkspaceClient
+
+    monkeypatch.setenv("DATABRICKS_HOST", "http://localhost")
+    monkeypatch.setenv("DATABRICKS_TOKEN", "test_token")
+    w = WorkspaceClient(config=Config())
+
     headers = {
         "Accept": "text/plain",
         "Content-Type": "application/json",


databricks/sdk/config.py
```diff
@@ -46,7 +46,10 @@ def __get__(self, cfg: "Config", owner):
         return cfg._inner.get(self.name, None)
 
     def __set__(self, cfg: "Config", value: any):
-        cfg._inner[self.name] = self.transform(value)
+        if value is None:
```
hectorcast-db, line 49: This is a bug fix. Before, this:
```
cfg.account_id = None
```
would actually set 
```
cfg.account_id = "None"
```

databricks/sdk/config.py
```diff
@@ -335,33 +344,25 @@ def _get_azure_environment_name(self) -> str:
...
-            for environment in ALL_ENVS:
-                if environment.cloud != Cloud.AZURE:
-                    continue
-                if environment.azure_environment.name != azure_env:
-                    continue
-                if environment.dns_zone.startswith(".dev") or environment.dns_zone.startswith(".staging"):
-                    continue
-                return environment
-        return get_environment_for_hostname(self.host)
+        if not self.experimental_is_unified_host:
```
hectorcast-db, line 347: We now resolve this once during new Config, not lazily. This ensures that we don't read the host after Config is created.

databricks/sdk/config.py
```diff
@@ -335,33 +344,25 @@ def _get_azure_environment_name(self) -> str:
...
+            # TODO: Remove this when making the unified mode the default.
+            self._resolve_environment()
+        return self.databricks_environment
 
     @property
     def is_azure(self) -> bool:
         if self.azure_workspace_resource_id:
             return True
-        return self.environment.cloud == Cloud.AZURE
+        return self.environment is not None and self.environment.cloud == Cloud.AZURE
```
hectorcast-db, line 357: Unified Hosts will return false on all 3 clouds.

databricks/sdk/config.py
```diff
@@ -525,12 +527,9 @@ def sql_http_path(self) -> Optional[str]:
             return None
         if self.cluster_id and self.warehouse_id:
             raise ValueError("cannot have both cluster_id and warehouse_id")
-        headers = self.authenticate()
```
hectorcast-db, line 528: Extracted to a separate `func`

PR #1257, opened 2026-02-11
Remove Cloud specific checks from the Python SDK

hectorcast-db:
This PR removes Cloud specific checks from the Python SDK.

In particular, it tries all authentication methods for all clouds, and filters based on the provided configuration fields.
This allows usage of Cloud authentication with any host such as SPOG and alias hosts.

## How is this tested?
* Added new unit tests. All tests use mocking to:
  - Simulate cloud-agnostic hosts (https://api.databricks.com)
  - Mock actual credential sources (avoid real Azure/Google API calls)
  - Verify headers are correctly generated
  - Confirm provider activation without cloud checks

*  This complements the existing integration tests and provides regression coverage for the removed cloud type validations.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Release v0.88.0
 
 ### New Features and Improvements
+* Remove cloud type restrictions from Azure/GCP credential providers. Azure and GCP authentication now works with any Databricks host when credentials are properly configured,enabling authentication against cloud-agnostic endpoints such as aliased hosts.
 
 ### Security
 


databricks/sdk/credentials_provider.py
```diff
@@ -814,7 +807,7 @@ def get_subscription(cfg: "Config") -> Optional[str]:
         return components[2]
 
 
-@credentials_strategy("azure-cli", ["is_azure"])
+@credentials_strategy("azure-cli", ["effective_azure_login_app_id"])
```
tanmay-db, line 810: Since `is_azure` checks for `azure_workspace_resource_id` and now we would be checking for `effective_azure_login_app_id`, could this be a breaking change?

hectorcast-db: the token is retrieved for a specific app id, just a few lines below:
```
    inner = token_source_for(cfg.effective_azure_login_app_id)
```
So on one side we do need this to make the call.
On another side, this only exists for Azure hosts (whether manually configured or automatically detected).
Unless I am missing something, they are equivalent

PR #1271, opened 2026-02-18
Dedup user agent

artjen:

**WHAT**: Make `useragent.with_extra()` idempotent by skipping duplicate `(key, value)` pairs that already exist in the global `_extra` list. This is a two-line change — an `if (key, value) in _extra: return` guard before the append.

**WHY**: LiteLLM's Databricks provider calls `useragent.with_partner("litellm")` on every LLM completion request (inside `_get_databricks_credentials()` in `litellm/llms/databricks/common_utils.py`). Because `with_partner()` delegates to `with_extra()`, which unconditionally appends to the module-level global `_extra` list, the User-Agent header grows by ~17 bytes per request and is never cleaned up.

In long-lived processes like model serving containers, this causes the User-Agent header to grow to tens of kilobytes after sustained traffic.

The fix is intentionally on the SDK side rather than in litellm because:
1. It protects against the same class of bug from any caller, not just litellm
2. The `_extra` list is typically <10 entries, so the `in` membership check is negligible cost
3. There is no legitimate use case for duplicate `(key, value)` pairs in the User-Agent — different values for the same key (e.g. `partner/litellm` + `partner/dspy`) are still allowed

## How is this tested?

Three new unit tests added to `tests/test_user_agent.py`:

- **`test_with_extra_is_idempotent`** — Calls `with_extra("partner", "litellm")` three times and asserts `partner/litellm` appears exactly once in the User-Agent string
- **`test_with_partner_is_idempotent`** — Calls `with_partner("litellm")` 100 times (simulating a long-lived serving container) and asserts `partner/litellm` appears exactly once
- **`test_with_extra_different_values_still_allowed`** — Calls `with_extra()` with two different values for the same key and asserts both are present, confirming we don't over-deduplicate

All existing tests continue to pass, including `test_user_agent_with_partner` which registers two different partners and asserts both appear.


tests/test_user_agent.py
```diff
@@ -45,6 +45,31 @@ def test_user_agent_with_partner(user_agent):
...
+    user_agent.with_extra("partner", "litellm")
+    assert user_agent.to_string().count("partner/litellm") == 1
+
+
+@pytest.mark.xdist_group(name="user_agent")
+def test_with_partner_is_idempotent(user_agent):
+    for _ in range(100):
+        user_agent.with_partner("litellm")
+    ua = user_agent.to_string()
+    assert ua.count("partner/litellm") == 1
```
renaudhartert-db: This is covered by the test above.

```suggestion
```

tests/test_user_agent.py
```diff
@@ -45,6 +45,31 @@ def test_user_agent_with_partner(user_agent):
     assert "partner/differenttest" in user_agent.to_string()
 
 
+@pytest.mark.xdist_group(name="user_agent")
+def test_with_extra_is_idempotent(user_agent):
+    user_agent.with_extra("partner", "litellm")
+    user_agent.with_extra("partner", "litellm")
+    user_agent.with_extra("partner", "litellm")
+    assert user_agent.to_string().count("partner/litellm") == 1
```
renaudhartert-db: Let's keep the code "anonymous"

```suggestion
    user_agent.with_extra("foo", "bar")
    user_agent.with_extra("foo", "bar")
    user_agent.with_extra("foo", "bar")
    assert user_agent.to_string().count("foo/bar") == 1
```

tests/test_user_agent.py
```diff
@@ -45,6 +45,31 @@ def test_user_agent_with_partner(user_agent):
...
+    assert ua.count("partner/litellm") == 1
+
+
+@pytest.mark.xdist_group(name="user_agent")
+def test_with_extra_different_values_still_allowed(user_agent):
+    user_agent.with_extra("partner", "litellm")
+    user_agent.with_extra("partner", "dspy")
+    ua = user_agent.to_string()
+    assert "partner/litellm" in ua
+    assert "partner/dspy" in ua
```
renaudhartert-db:
```suggestion
def test_with_extra_different_values_still_allowed(user_agent):
    user_agent.with_extra("foo", "bar")
    user_agent.with_extra("foo", "wiz")
    ua = user_agent.to_string()
    assert "foo/bar" in ua
    assert "foo/wiz" in ua
```

PR #1394, opened 2026-04-20
Expand AI agent detection: Goose, Amp, Augment, Copilot (VS Code), Kiro, Windsurf

simonfaltum:
## Why

We identify which AI agent is driving the SDK via the `agent/<name>` user-agent segment. The current list covers 9 agents. This adds 6 more and honors the emerging `AGENT=<name>` standard from agents.md so we can see traffic from agents we haven't individually listed yet.

## Changes

Before: presence-only matching on a fixed list of env vars; multi-match returns empty.

Now: each agent record has a list of matchers (env var plus optional exact value), an agent fires if any matcher fires, ambiguity is judged by unique product (not raw matcher count), and an `AGENT=<anything>` fallback reports `unknown` when no specific entry matched.

Implementation:
- `databricks/sdk/useragent.py`: replaced the `_KNOWN_AGENTS` dict with a list of `_AgentRecord` dataclasses each holding a `product` and a list of `(env_var, value)` matchers. Rewrote `agent_provider()` to count unique agents matched, handle the single/multi/zero cases, and fall back to `unknown` when `AGENT` is set non-empty but no specific agent matched. Cached result (`_agent_provider`) still uses the `None` vs `""` sentinel pattern.
- `tests/test_user_agent.py`: added coverage for each new agent (goose, amp, augment, copilot-vscode, kiro, windsurf), both Goose signals together (not ambiguous), both Amp signals together (not ambiguous), unknown fallback, empty `AGENT` does not trigger fallback, and cross-agent ambiguity (`AGENT=goose` + `CLAUDECODE=1`).
- `NEXT_CHANGELOG.md`: entry under "Internal Changes".

New detections: Goose (`GOOSE_TERMINAL` or `AGENT=goose`), Amp (`AMP_CURRENT_THREAD_ID` or `AGENT=amp`), Augment (`AUGMENT_AGENT`), VS Code Copilot (`COPILOT_MODEL`), Kiro (`KIRO`), Windsurf (`WINDSURF_AGENT`).

Parallel PRs with identical behavior are being opened in `databricks-sdk-go` and `databricks-sdk-java`.

## Test plan

- [x] `python3 -m pytest tests/test_user_agent.py -v` passes (37 tests)
- [x] `make fmt` leaves the three modified files clean
- [x] Each new product detected when its primary env var is set
- [x] `AGENT=goose` alone returns `goose`
- [x] `GOOSE_TERMINAL=1` + `AGENT=goose` returns `goose` (not ambiguity)
- [x] `AMP_CURRENT_THREAD_ID` + `AGENT=amp` returns `amp` (not ambiguity)
- [x] `AGENT=someweirdthing` returns `unknown`
- [x] `AGENT=""` returns `""` (fallback only fires on non-empty value)
- [x] Two distinct agents set simultaneously return `""`
- [x] `AGENT=goose` + `CLAUDECODE=1` returns `""` (cross-agent ambiguity)

### DIFF
--- a/tests/test_user_agent.py
+++ b/tests/test_user_agent.py
@@ -294,11 +294,22 @@ def test_agent_provider_agent_empty_string(clean_useragent_env):
 
 
 def test_agent_provider_multiple_agents(clean_useragent_env):
+    # Nested agents (e.g. Claude Code spawning a Cursor CLI subagent) set
+    # multiple explicit matchers on the same process.
     os.environ["CLAUDECODE"] = "1"
     os.environ["CURSOR_AGENT"] = "1"
     from databricks.sdk import useragent
 
-    assert useragent.agent_provider() == ""
+    assert useragent.agent_provider() == "multiple"
+
+
+def test_agent_provider_three_stacked_agents(clean_useragent_env):
+    os.environ["CLAUDECODE"] = "1"
+    os.environ["CURSOR_AGENT"] = "1"
+    os.environ["AUGMENT_AGENT"] = "1"
+    from databricks.sdk import useragent
+
+    assert useragent.agent_provider() == "multiple"
 
 
 def test_agent_provider_explicit_wins_over_agent_naming_different_product(clean_useragent_env):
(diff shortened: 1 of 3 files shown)


tests/test_user_agent.py
```diff
@@ -181,6 +181,118 @@ def test_agent_provider_openclaw(clean_useragent_env):
...
+
+    assert useragent.agent_provider() == "claude-code"
+
+
+def test_agent_provider_agent_empty_string(clean_useragent_env):
+    # AGENT="" (empty) should NOT trigger the fallback.
+    os.environ["AGENT"] = ""
+    from databricks.sdk import useragent
+
+    assert useragent.agent_provider() == ""
```
mihaimitrea-db, line 293: Ideally we would have these tests as a single table driven test. It's easier to read and easier to expand if more models are added in the future.

mihaimitrea-db (approved):

PR #1436, opened 2026-05-20
Use X-Databricks-Workspace-Id for workspace routing

Divyansh-db:
## Summary

Workspace-scoped API **requests** now identify the target workspace via the `X-Databricks-Workspace-Id` header instead of `X-Databricks-Org-Id`. The value source is unchanged: it still comes from `Config.workspace_id` (or the `DATABRICKS_WORKSPACE_ID` environment variable). Response-side reads of `/api/2.0/preview/scim/v2/Me` continue to read the workspace ID from the legacy `X-Databricks-Org-Id` response header — this PR is **outbound-only**.

## Why

On unified Databricks hosts that serve multiple workspaces, the SDK sends a routing header on every workspace-scoped HTTP request so the gateway can dispatch it to the correct workspace. Until now that header was `X-Databricks-Org-Id`.

The Databricks platform is consolidating workspace addressing onto a single header, `X-Databricks-Workspace-Id`, which accepts a broader range of workspace identifier formats and replaces the older single-purpose channels going forward. The previous header continues to be accepted by the platform, so older SDK versions keep working, but new SDK versions should use the new name.

This PR is the Python SDK's part of that migration.

### Interface changes

None. `Config.workspace_id` keeps the same field name, type (`str`), and environment variable (`DATABRICKS_WORKSPACE_ID`). The accompanying comment is widened to note that the value may now be either a classic numeric workspace ID or another workspace identifier format that the server understands.

### Behavioral changes

- Workspace-scoped **requests** no longer send `X-Databricks-Org-Id`. They now send `X-Databricks-Workspace-Id` (with the same value, gated on `Config.workspace_id` being non-empty). Account-scoped requests are unaffected.
- **Response-side reads are unchanged.** The server still emits the workspace ID in the legacy `X-Databricks-Org-Id` response header on `/api/2.0/preview/scim/v2/Me`, so `WorkspaceClient.get_workspace_id()`, the storage-proxy workspace probe in `FilesExt._resolve_workspace_id()`, and the SQL HTTP path resolver in `Config._workspace_id_from_jdbc_path()` all continue to read the old header name.

### Internal changes

- Regenerated `databricks/sdk/service/*.py` (including the newly-added `bundle.py`), `databricks/sdk/__init__.py`, and the generated test fixtures under `tests/databricks/sdk/service/` from the corresponding generator template change. Pure mechanical literal swap on workspace-scoped request emission.
- Updated the hand-written sites that emit the same header on workspace-scoped calls but are not covered by the generator: `databricks/sdk/mixins/workspace.py` (upload/download), `databricks/sdk/mixins/sharing.py` (list shares).
- Widened the docstring on `Config.workspace_id` to note the new accepted formats.

### Test changes

- Renamed two request-side unit tests in `tests/test_config.py` so the names reflect the new request-side header (`test_*_org_id_header_*` → `test_*_workspace_id_header_*`).
- Upgraded the third renamed test into a direct unit test on `WorkspaceClient.get_workspace_id()`: `test_get_workspace_id_reads_org_id_response_header_when_config_missing_workspace_id` now mocks `/api/2.0/preview/scim/v2/Me` returning an `X-Databricks-Org-Id` response header and asserts both (a) the SDK does not send the new request header when `Config.workspace_id` is empty, and (b) the SDK correctly parses the legacy response header.
- Added a wire-level integration test `test_spog_unified_host_sends_workspace_id_header` in `tests/integration/test_unified_profile.py` that mounts a transport-level capture adapter on the SDK's `requests.Session` and probes `/api/2.0/preview/scim/v2/Me` against a real unified host, asserting (1) the request carries `X-Databricks-Workspace-Id`, (2) the request does not carry the legacy `X-Databricks-Org-Id`, and (3) the response still echoes `X-Databricks-Org-Id`. Mirrors the Go SDK's equivalent acceptance test.

## How is this tested?

- `make test` passes locally (2073 passed, 3 skipped).
- Unit coverage:
  - The two renamed request-side tests in `tests/test_config.py` cover request-header emission on workspace-scoped vs account-scoped calls on unified hosts.
  - `test_get_workspace_id_reads_org_id_response_header_when_config_missing_workspace_id` is a direct unit test on `WorkspaceClient.get_workspace_id()` covering both directions (no new request header + legacy response header parsed correctly) on the SCIM /Me probe.
  - The storage-proxy / SCIM-`/Me` response-parsing paths in `mixins/files.py` are exercised by `tests/test_files.py` (mock fixtures continue to emit `X-Databricks-Org-Id`, matching the unchanged response-side behavior).
- Integration coverage:
  - `tests/integration/test_unified_profile.py::test_spog_unified_host_sends_workspace_id_header` verifies wire-level header behavior against a real unified host (runs in the OAuth M2M-equipped matrices).
- Spot-checked the regenerated diff: every line change in `databricks/sdk/service/*.py` is exclusively the header literal swap; no collateral changes.


hectorcast-db (approved):

hectorcast-db (commented):
LGTM

PR #1460, opened 2026-06-02
Cache tokens minted by DatabricksOidcTokenSource

hectorcast-db:
## What

`DatabricksOidcTokenSource` (Workload Identity Federation / account-wide token federation) minted a fresh `/oidc/v1/token` exchange on **every** authenticated API call. Its `token()` always called `_exchange_id_token()`, which constructed a brand-new `oauth.ClientCredentials` each time — discarding the `Refreshable` caching/locking/refresh machinery already built into it.

This change makes `DatabricksOidcTokenSource` extend `oauth.Refreshable` and moves the exchange logic into `refresh()`, so the minted token is cached and reused until it is stale or expired (mirroring the existing M2M/PAT flows). Each refresh re-fetches the ID token from the `id_token_source`, so short-lived **rotating** ID tokens are handled transparently.

## Why

The previous behaviour:
- added an extra RTT to every request,
- amplified transient federation-policy errors by the number of API calls,
- and hit OIDC token-endpoint rate limits.

A real-world 470 MiB Volume upload minted dozens of tokens in under 30s, with bursts of 6–7 concurrent mints.

## Tests

Added unit tests in `tests/test_oidc.py`:
- token is minted once and reused across calls (caching),
- an expired token triggers a re-exchange that fetches a **fresh** ID token (rotation),
- missing host still raises `ValueError`.

All existing auth tests pass (`test_oauth`, `test_oidc`, `test_credentials_provider`, `test_oidc_token_supplier`). Added a `NEXT_CHANGELOG.md` Bug Fixes entry.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bug Fixes
 
+* Cache tokens minted by `DatabricksOidcTokenSource` (Workload Identity Federation / account-wide token federation). Previously a fresh `/oidc/v1/token` exchange was performed on every authenticated API call, adding latency, amplifying transient federation-policy errors, and hitting OIDC token-endpoint rate limits. The token source now reuses the cached token until it is stale or expired, fetching a fresh ID token on each refresh to support rotation.
+
 ### Documentation
 
 ### Breaking Changes
--- a/databricks/sdk/oidc.py
+++ b/databricks/sdk/oidc.py
@@ -123,8 +123,14 @@ class FileIdTokenSource(IdTokenSource):
         return IdToken(jwt=token)
 
 
-class DatabricksOidcTokenSource(oauth.TokenSource):
+class DatabricksOidcTokenSource(oauth.Refreshable):
     """A TokenSource which exchanges a token using Workload Identity Federation.
+
+    The exchanged token is cached and reused across calls. It is only refreshed
+    when it is stale or expired, mirroring the M2M/PAT auth flows. This avoids
+    minting a fresh token on every authenticated API call. Each refresh fetches a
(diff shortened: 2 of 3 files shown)


renaudhartert-db (commented):
Interesting, I thought we reimplemented the token cache to be a wrapper like in other SDKs rather than a class to be extended. Wasn't that the case?

hectorcast-db:
> Interesting, I thought we reimplemented the token cache to be a wrapper like in other SDKs rather than a class to be extended. Wasn't that the case?

Looking at the Git history, Python token cache predates Go token cache by 2 years (it's actually part of the "initial" [commit](https://github.com/databricks/databricks-sdk-py/commit/b4acecfa)). We just updated the Python token cache to support async token refresh but didn't reimplemented the full cache as a wrapper.

renaudhartert-db (approved):
LGTM. Clean, minimal fix that correctly moves the exchange into `refresh()` so the base `Refreshable` handles caching, locking, and staleness/expiry; matches the existing M2M/PAT flows.

Verified against the base class:
- Missing-host `ValueError` still surfaces synchronously (first call sees an empty/EXPIRED token and takes the blocking path).
- ID-token rotation works since each `refresh()` re-fetches from `id_token_source`.
- Concurrent callers collapse to a single mint via the lock + `_is_refreshing` dedupe, addressing the burst-of-concurrent-mints case.
- `disable_async` is correctly delegated to the base `__init__`; in practice it's wired from `cfg.disable_async_token_refresh` like the other flows.

Tests cover caching, expiry-triggered re-exchange with a fresh ID token, and missing-host; the expired case stays on the deterministic blocking path. All pass.

PR #1461, opened 2026-06-03
fix(dbutils): accept pathlib.Path arguments in fs methods

### LINKED ISSUE #1459: [ISSUE] DBUtils does not support Path objects.
**Problem Statement**
When I pass in a `pathlib.Path` to `DBUtils` functions they fail, despite pathlib being the default way to do paths in Python. E.g., `dbutils.fs.ls(Path("my/cool/folder"))`

**Proposed Solution**
Support `Path` objects as input to `dbutils` functions, could be done by just casting to string.

**Additional Context**
I don't know if the actual dbutils code is maintained here, but you at least have the Workspace Client implementation so figured you guys would know.

nileshpatil6:
Fixes #1459

`dbutils.fs` methods only accepted plain strings, so passing a `pathlib.Path` failed even though Path is idiomatic in modern Python.

**What changed**

Added a small `_as_str` helper in `databricks/sdk/dbutils.py` that converts a `Path` to `str` and leaves strings untouched. Called it at the entry of every `_FsUtil` method that takes a path argument (`ls`, `cp`, `mv`, `rm`, `mkdirs`, `head`, `put`, `mount`, `unmount`, `updateMount`).

**Tests**

Added 7 unit tests (one per patched method) that pass a `Path` object and assert the underlying API is still called with a plain string.

Signed-off-by: nileshpatil6 <technil6436@gmail.com>


GodNooNoo:
@nileshpatil6 any updates here?

nileshpatil6:
Hey, sorry for the delay. The branch had gone stale, upstream removed NEXT_CHANGELOG.md (and its workflow) in the v0.117.0 release, which is what caused the merge conflict. I've merged the latest main and dropped the changelog hunk since that whole mechanism is gone now. The diff is back to just the two files: dbutils.py and its tests. Reran the full tests/test_dbutils.py suite locally, 26/26 pass, and ruff format/lint (pinned 0.5.6) are clean. Worth noting per the updated CONTRIBUTING.md this repo is now a read-only mirror, so even once approved a maintainer re-applies the change internally and it ships with a release rather than merging here directly.

PR #1464, opened 2026-06-08
Fall back to remote runtime on Spark Connect when the legacy namespace is unavailable

### LINKED ISSUE #1463: WorkspaceClient construction fails on Spark Connect clusters: CONTEXT_UNAVAILABLE_FOR_REMOTE_CLIENT when importing databricks.sdk.runtime
### Description

On a Databricks shared-access-mode (Spark Connect) cluster, simply *constructing* a `WorkspaceClient()` raises `CONTEXT_UNAVAILABLE_FOR_REMOTE_CLIENT` — before any API call is made. This surfaces in downstream tools that build a `WorkspaceClient` on such a cluster; for example dbt-databricks Python models (see [dbt-databricks#1252](https://github.com/databricks/dbt-databricks/issues/1252)).

### Root cause

`WorkspaceClient.__init__` eagerly runs `self._dbutils = _make_dbutils(self._config)`. On a cluster, `_make_dbutils` finds `dbruntime` importable, so it takes the runtime branch and does `from databricks.sdk.runtime import dbutils`.

Importing `databricks.sdk.runtime` runs module-level code that calls `UserNamespaceInitializer.getOrCreate().get_namespace_globals()`, which materializes a **legacy `SparkContext`**. Under Spark Connect that raises `CONTEXT_UNAVAILABLE_FOR_REMOTE_CLIENT` (a `pyspark.errors.PySparkRuntimeError`, **not** an `ImportError`), so the surrounding `except ImportError:` in `databricks/sdk/runtime/__init__.py` does not catch it and the error propagates out of the import — crashing the `WorkspaceClient` constructor.

```
databricks/sdk/__init__.py        WorkspaceClient.__init__ -> _make_dbutils(config)
databricks/sdk/__init__.py        _make_dbutils -> from databricks.sdk.runtime import dbutils   (dbruntime importable)
databricks/sdk/runtime/__init__.py  UserNamespaceInitializer.getOrCreate().get_namespace_globals()
                                  -> builds a legacy SparkContext
                                  -> raises [CONTEXT_UNAVAILABLE_FOR_REMOTE_CLIENT]   (not ImportError)
```

### Why a consumer-side workaround is awkward

Because the failure is in the constructor (transitively), a `try/except` around `ws.dbutils` access cannot rescue it — the object is never returned. There is also no `Config` flag or environment variable to skip the eager dbutils build. As a result, downstream tools can at max resort to monkey-patching the private `databricks.sdk._make_dbutils` symbol, which would be brittle.

### Expected behavior

Importing `databricks.sdk.runtime` — and therefore constructing a `WorkspaceClient` — on a Spark Connect runtime should not raise. The legacy `SparkContext` is unavailable in remote clients; the SDK already has a Spark Connect-compatible "OSS/remote" code path in `databricks/sdk/runtime/__init__.py` (it builds `spark` via `DatabricksSession` and `dbutils` via `RemoteDbUtils`), and the runtime import should fall back to it.

### Proposed fix

**Layer 1 (PR attached):** In `databricks.sdk.runtime`, treat a failure while materializing the legacy user namespace the same as "not in a classic runtime" — log a warning and fall back to the existing OSS/remote implementation instead of propagating. This is a small, local change in a hand-written file.

**Layer 2 (design discussion):** Make `WorkspaceClient.dbutils` lazy (a `@property` / `cached_property`) so consumers that never touch `.dbutils` (which is most programmatic callers, including dbt) never build it, sidestepping this path entirely. This touches the generated `databricks/sdk/__init__.py`, so it would need to be mirrored into the generator — hence raising it as a discussion rather than a direct PR.

### Related

- #986 — mirror-image symptom: importing the runtime package eagerly builds `RemoteDbUtils()` and fails auth off-runtime. Same import-time side-effect class of problem.

### Environment

- databricks-sdk-py: reproduces on current versions including `main`.
- Cluster: Databricks shared-access-mode / Spark Connect (`SPARK_CONNECT_MODE_ENABLED`).

sd-db:
## Summary

Importing `databricks.sdk.runtime` on a Spark Connect runtime no longer raises `CONTEXT_UNAVAILABLE_FOR_REMOTE_CLIENT`. When the legacy user namespace can't be materialized, the import now logs a warning and falls back to the existing Spark Connect-compatible remote implementation. This fixes the upstream crash behind [dbt-databricks#1252](https://github.com/databricks/dbt-databricks/issues/1252).

Fixes #1463

## Why

`WorkspaceClient.__init__` eagerly builds dbutils via `_make_dbutils`, which on a cluster imports `databricks.sdk.runtime`. That import calls `UserNamespaceInitializer.getOrCreate().get_namespace_globals()`, materializing a **legacy `SparkContext`**. On a shared-access-mode (Spark Connect) cluster this raises `CONTEXT_UNAVAILABLE_FOR_REMOTE_CLIENT` — a `pyspark.errors.PySparkRuntimeError`, **not** an `ImportError` — so the existing `except ImportError:` doesn't catch it and the error escapes the import, crashing `WorkspaceClient` construction before any API call.

Because the failure is in the constructor, consumers can't guard it with a `try/except` around `ws.dbutils`, and there's no config/env switch to skip the eager build — so downstream tools resort to monkey-patching the private `_make_dbutils`. The `except ImportError` branch is already the Spark Connect-compatible path (it builds `spark` via `DatabricksSession` and `dbutils` via `RemoteDbUtils`), so this PR routes the materialization failure there.

See #1463 for the full analysis. A complementary follow-up (making `WorkspaceClient.dbutils` lazy) is noted there as a separate discussion since it touches generated code.

### Interface changes

None.

### Behavioral changes

On a Spark Connect runtime, importing `databricks.sdk.runtime` now logs a `WARNING` and uses the remote implementation instead of raising at import time. When `dbruntime` is absent, or when the namespace materializes successfully, behavior is unchanged.

### Internal changes

Restructured the runtime-namespace block in `databricks/sdk/runtime/__init__.py` into a single `try` with sibling `except ImportError` (not a classic runtime) and `except Exception` (Spark Connect / `CONTEXT_UNAVAILABLE_FOR_REMOTE_CLIENT`, logged) clauses, plus an `if not _use_runtime_namespace:` guard over the existing — unchanged — OSS/remote block.

## How is this tested?

New `tests/test_runtime.py` simulates a Spark Connect runtime by injecting a fake `dbruntime` whose `get_namespace_globals()` raises `CONTEXT_UNAVAILABLE_FOR_REMOTE_CLIENT`, then asserts that (a) reloading `databricks.sdk.runtime` survives and falls back (`is_local_implementation is True`, `dbutils is not None`) and (b) `WorkspaceClient(config=…)` constructs without raising — the direct reproduction of the reported failure. Verified red→green (both tests fail with the change reverted, pass with it). The full unit suite shows no regressions introduced by this change.


Divyansh-db:
Hi @sd-db — thanks for the detailed analysis and the fix. Carrying this forward upstream.

Tests don't run on forked PRs, so the cleanest move was to land the same commit on an upstream branch. Opened #1469 with your commit preserved.

Closing this in favor of #1469. Thanks again for the contribution!

PR #1481, opened 2026-06-23
Send an empty JSON body for parameterless POST/PUT/PATCH requests

tejaskochar-db:
## Summary

Send an empty JSON object as the body for `POST`/`PUT`/`PATCH` requests that declare `Content-Type: application/json` but have no body to serialize, so the request is well-formed on the wire instead of being sent with no body at all.

## Why

API methods for resource actions that take no parameters — for example `WorkspaceClient.knowledge_assistants.sync_knowledge_sources(...)`, which issues `POST .../knowledge-sources:sync` — set `Content-Type: application/json` but pass no `body` to `_BaseClient.do`.

`body` defaults to `None`, and `_perform` calls `requests` with `json=None`. `requests` treats `json=None` as *no body*: the request goes out with `Content-Length: 0` while still advertising `Content-Type: application/json`. Some Databricks API gateways require a JSON body (even an empty `{}`) for these endpoints and reject the bodyless request with `400 Bad Request`.

The same operation succeeds when a body is supplied — calling `w.api_client.do("POST", ".../knowledge-sources:sync", body={})` directly works. The proof is in how `requests` prepares the body:

| call          | wire body | Content-Length |
|---------------|-----------|----------------|
| `json=None`   | `None`    | 0              |
| `json={}`     | `b"{}"`   | 2              |

This affects every generated method that declares a JSON content type but has no request fields (resource actions such as `:sync`, `start`, `stop`, `cancel`, `undelete`, …) — 20 such methods in the SDK today, and any added later.

### Interface changes

None.

### Behavioral changes

`_BaseClient.do` now defaults `body` to `{}` when **all** of the following hold: the method is `POST`/`PUT`/`PATCH`, `Content-Type` is `application/json`, and no `body`, `data`, or `files` is supplied. These requests previously went out with no body; they now carry `{}`.

`GET`/`DELETE`/`HEAD`, requests that already carry a body, and requests that do not declare `application/json` (e.g. streamed `application/octet-stream` uploads, or actions that intentionally send no body) are unchanged.

This mirrors the Go SDK, which marshals an empty request struct to `{}` and so has always sent a body for these endpoints.

### Internal changes

None.

## How is this tested?

Unit tests in `tests/test_base_client.py` cover the positive case parametrized over `POST`/`PUT`/`PATCH` (now sends `{}`) and the boundaries that must stay unchanged: a `POST` without a JSON content type sends no body; a `GET` is never given a body; an explicit `files={}` is left alone (no body injected); and an explicitly supplied falsy body such as `[]` is sent verbatim. Each test was confirmed to fail when the corresponding production logic is broken. Full `test_base_client.py` suite passes.

---
This pull request and its description were written by Isaac.

### DIFF
--- a/databricks/sdk/_base_client.py
+++ b/databricks/sdk/_base_client.py
@@ -162,6 +162,24 @@ class _BaseClient:
             headers = {}
         headers["User-Agent"] = self._user_agent_base
 
+        # A request that advertises a JSON content type must still send a JSON body. Resource
+        # actions that take no parameters (e.g. ".../knowledge-sources:sync") set
+        # Content-Type: application/json but have no fields to serialize, so the caller passes
+        # no body. requests turns json=None into no request body at all, which some Databricks
+        # API gateways reject. Default the body to an empty object so the request is well-formed.
+        # The Content-Type comparison is intentionally an exact match: every generated call site
+        # emits this literal, so matching it narrowly avoids touching callers that set a variant.
+        # `data`/`files` are checked with `is None` because an explicitly supplied payload — even
+        # an empty one — means the caller is sending its own body, not requesting the default.
+        if (
+            body is None
+            and data is None
+            and files is None
+            and method in ("POST", "PUT", "PATCH")
+            and headers.get("Content-Type") == "application/json"
+        ):
+            body = {}
+
         # Wrap strings and bytes in a seekable stream so that we can rewind them.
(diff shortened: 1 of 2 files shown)


tejaskochar-db:
Closing this manual draft. This change will be delivered through the regular automated SDK update instead.

PR #1497, opened 2026-07-24
fix: handle non-string request bodies in error-logging round-trip dump

### LINKED ISSUE #1489: [ISSUE] `get_api_error` fails with `TypeError: object of type '_io.BufferedReader' has no len()¨`
**Description**

When the HTTP PUT request body is streamed from a local file, `request.body` is not a string but a `_io.BufferedReader`.

In case of API error, `_BaseClient._error_parser.get_api_error(response)` will try to return the error and the associated query.

https://github.com/databricks/databricks-sdk-py/blob/be7acc03728c5adf571f60d49ace4fd707d48f8b/databricks/sdk/errors/parser.py#L93-L96

Unfortunately, `response.request.body` is still a BufferedReader, not necessarily a string. When this `body` is passed to the logger, it fails

https://github.com/databricks/databricks-sdk-py/blob/be7acc03728c5adf571f60d49ace4fd707d48f8b/databricks/sdk/logger/round_trip_logger.py#L46-L47

eventually calls `self._redacted_dump("> ", request.body)` but the latter is not meant to handle streams

Full strack trace : 

```bash
Traceback (most recent call last):
  File "C:\_dev\my_proj\.venv\Lib\site-packages\dvc_objects\fs\generic.py", line 350, in transfer
    _try_links(
  File "C:\_dev\my_proj\.venv\Lib\site-packages\dvc_objects\fs\generic.py", line 281, in _try_links
    return copy(
           ^^^^^
  File "C:\_dev\my_proj\.venv\Lib\site-packages\dvc_objects\fs\generic.py", line 88, in copy
    return _put(
           ^^^^^
  File "C:\_dev\my_proj\.venv\Lib\site-packages\dvc_objects\fs\generic.py", line 161, in _put
    _put_one(from_paths[0], to_paths[0])
  File "C:\_dev\my_proj\.venv\Lib\site-packages\dvc_objects\fs\generic.py", line 151, in _put_one
    return to_fs.put_file(
           ^^^^^^^^^^^^^^^
  File "C:\_dev\my_proj\.venv\Lib\site-packages\dvc_objects\fs\base.py", line 673, in put_file
    self.fs.put_file(os.fspath(from_file), to_info, callback=callback, **kwargs)
  File "C:\_dev\my_proj\.venv\Lib\site-packages\dvc_databricks\filesystem.py", line 308, in put_file
    self._client.files.upload(rpath, fh, overwrite=True)
  File "C:\_dev\my_proj\.venv\Lib\site-packages\databricks\sdk\mixins\files.py", line 1430, in upload
    self._upload_single_thread_with_known_size(ctx, contents)
  File "C:\_dev\my_proj\.venv\Lib\site-packages\databricks\sdk\mixins\files.py", line 1513, in _upload_single_thread_with_known_size
    return self._single_thread_single_shot_upload(ctx, contents)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\_dev\my_proj\.venv\Lib\site-packages\databricks\sdk\mixins\files.py", line 1521, in _single_thread_single_shot_upload
    return super().upload(file_path=ctx.target_path, contents=contents, overwrite=ctx.overwrite)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\_dev\my_proj\.venv\Lib\site-packages\databricks\sdk\service\files.py", line 1027, in upload
    self._api.do(
  File "C:\_dev\my_proj\.venv\Lib\site-packages\databricks\sdk\core.py", line 95, in do
    return self._api_client.do(
           ^^^^^^^^^^^^^^^^^^^^
  File "C:\_dev\my_proj\.venv\Lib\site-packages\databricks\sdk\_base_client.py", line 197, in do
    response = call(
               ^^^^^
  File "C:\_dev\my_proj\.venv\Lib\site-packages\databricks\sdk\retries.py", line 59, in wrapper
    raise err
  File "C:\_dev\my_proj\.venv\Lib\site-packages\databricks\sdk\retries.py", line 38, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "C:\_dev\my_proj\.venv\Lib\site-packages\databricks\sdk\_base_client.py", line 297, in _perform
    error = self._error_parser.get_api_error(response)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\_dev\my_proj\.venv\Lib\site-packages\databricks\sdk\errors\parser.py", line 95, in get_api_error
    {"message": "unable to parse response. " + _unknown_error(response, self._debug_headers)},
                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\_dev\my_proj\.venv\Lib\site-packages\databricks\sdk\errors\parser.py", line 45, in _unknown_error
    request_log = RoundTrip(response, debug_headers=debug_headers, debug_truncate_bytes=10 * 1024).generate()
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\_dev\my_proj\.venv\Lib\site-packages\databricks\sdk\logger\round_trip_logger.py", line 47, in generate
    sb.append("> [raw stream]" if self._raw else self._redacted_dump("> ", request.body))
                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\_dev\my_proj\.venv\Lib\site-packages\databricks\sdk\logger\round_trip_logger.py", line 113, in _redacted_dump
    if len(body) == 0:
       ^^^^^^^^^
TypeError: object of type '_io.BufferedReader' has no len()
```

A suggested fix could be : 

```python
    def _redacted_dump(self, prefix: str, body: str) -> str:
        if isinstance(body, io.IOBase):
            # Read the stream into a string-like
            body.seek(0)
            body = body.read().decode("utf-8")
        
        if len(body) == 0:
            return ""
```

mittalpk:
## Summary

`RoundTrip.generate()` (`databricks/sdk/logger/round_trip_logger.py`) passes `request.body` directly to `_redacted_dump()`, which assumes a `str`. This crashes in two ways, both inside the "unable to parse response" debug-log path itself — so the crash masks the actual API error behind an unrelated Python traceback:

1. **Streamed upload** (a file-like object passed as `data=` to `requests`, e.g. `dbutils.files.upload` from a local file): `requests` leaves `request.body` as the original file object rather than reading it. `_redacted_dump`'s `len(body)` call then raises `TypeError: object of type '...' has no len()`.
2. **Non-UTF8 bytes body**: once JSON parsing fails, the fallback path calls `body.encode(...)` on an already-`bytes` value, raising `AttributeError: 'bytes' object has no attribute 'encode'`.

Fixes #1489 (the streamed-upload crash); the non-UTF8-bytes crash is the same root cause (an assumption that `request.body` is always `str`) so it's fixed and tested alongside it rather than left for a follow-up.

## Fix

Added `_request_body_str()` to normalize `request.body` into a string before it reaches `_redacted_dump`:
- `str` passes through unchanged.
- `bytes`/`bytearray` decode with `errors="replace"` (matching the existing pattern already used for `response.content` a few lines below).
- Anything else (a stream/file-like object) is represented with a fixed `"[stream body]"` placeholder rather than read — reading it here would consume the stream (which may already be at an unknown position after the actual send) with no guarantee its content is UTF-8 text.

## Test plan

- Added two cases to `tests/test_errors.py::test_get_api_error` covering both crash scenarios (streamed body via `io.BytesIO`, non-UTF8 `bytes` body), extending the existing `fake_raw_response` helper with an optional `request_body` parameter that's passed through to `requests.Request(..., data=...).prepare()` — this reproduces exactly how `requests` represents each body type in production, rather than hand-constructing a fake object.
- Both new tests verified to fail against pre-fix code (`git stash` the fix in `round_trip_logger.py`, rerun — one raises `TypeError`, the other `AttributeError`, matching the two reported/found crash shapes) and pass post-fix.
- `pytest tests/test_errors.py`: 52 passed.
- `pytest tests/` (excluding `tests/integration`): 2109 passed; the only failures (`test_open_ai_mixin.py`, 7 tests) are pre-existing in this environment due to the optional `openai`/`langchain` packages not being installed, unrelated to this change — confirmed by running that file alone and seeing the same `ModuleNotFoundError` on unmodified `main`.
- `ruff check` / `ruff format --check`: clean on both changed files.

### DIFF
--- a/databricks/sdk/logger/round_trip_logger.py
+++ b/databricks/sdk/logger/round_trip_logger.py
@@ -44,7 +44,9 @@ class RoundTrip:
             for k, v in request.headers.items():
                 sb.append(f"> * {k}: {self._only_n_bytes(v, self._debug_truncate_bytes)}")
         if request.body:
-            sb.append("> [raw stream]" if self._raw else self._redacted_dump("> ", request.body))
+            sb.append(
+                "> [raw stream]" if self._raw else self._redacted_dump("> ", self._request_body_str(request.body))
+            )
         sb.append(f"< {self._response.status_code} {self._response.reason}")
         if self._raw and self._response.headers.get("Content-Type", None) != "application/json":
             # Raw streams with `Transfer-Encoding: chunked` do not have `Content-Type` header
@@ -53,6 +55,22 @@ class RoundTrip:
             decoded = self._response.content.decode("utf-8", errors="replace")
             sb.append(self._redacted_dump("< ", decoded))
         return "\n".join(sb)
+
+    @staticmethod
+    def _request_body_str(body: Any) -> str:
+        """Normalize a `requests.PreparedRequest.body` into a string safe to hand to `_redacted_dump`.
+
+        The body is usually `str` or `bytes`, but for a streamed upload (e.g. a file object passed as
+        `data=`) `requests` leaves it as the original file-like object. Reading it here for logging
+        purposes would consume the stream (which may already be at an unknown position after the actual
+        send) and isn't guaranteed to be UTF-8, so it's represented with a fixed placeholder instead of
+        being read.
+        """
+        if isinstance(body, str):
+            return body
+        if isinstance(body, (bytes, bytearray)):
+            return bytes(body).decode("utf-8", errors="replace")
+        return "[stream body]"
 
     @staticmethod
     def _mask(m: Dict[str, any]):
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -1,4 +1,5 @@
 import http.client
+import io
 import json
 from dataclasses import dataclass, field
 from typing import Any, List, Optional
@@ -24,13 +25,14 @@ def fake_raw_response(
     status_code: int,
     response_body: bytes,
     path: Optional[str] = None,
+    request_body: Any = None,
 ) -> requests.Response:
     resp = requests.Response()
     resp.status_code = status_code
     resp.reason = http.client.responses.get(status_code, "")
     if path is None:
         path = "/api/2.0/service"
-    resp.request = requests.Request(method, f"https://databricks.com{path}").prepare()
+    resp.request = requests.Request(method, f"https://databricks.com{path}", data=request_body).prepare()
     resp._content = response_body
     return resp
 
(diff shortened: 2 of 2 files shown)


mittalpk:
Bumping this — let me know if anything's needed on my end to move it forward.

PR #1506, opened 2026-07-30
test: fmt experiment (before, current main) — do not merge

Divyansh-db:
Throwaway PR to observe the `fmt` CI check on the current-spec regenerated SDK surface **before** an import-sorting fix, based on the same `main` as the paired after PR. Expected: `fmt` fails (unsorted imports). Will be closed after inspection.


Divyansh-db:
fmt experiment complete (before-fix: fmt fails as expected). Throwaway PR — closing.

PR #1545, opened 2026-08-23
Improve missing cluster ID guidance

### LINKED ISSUE #51: Make error handling more user friendly for common accidental mistakes by showing resolution steps
This is mostly a nice to have. Example -- running dbutils.help() without setting up cluster throws: 
`databricks.sdk.core.DatabricksError: Missing required field: cluster_id`

We could add something along the lines of how to setup cluster, point to the relevant docs.

Full error stack:
image

TangoEnSkai:
## Summary

Improves the error shown when a remote DBUtils operation requires a cluster but no cluster ID is configured.

Fixes #51.

## Why

The existing error says that `cluster_id` is required, but it does not tell users how to provide it. Users can otherwise miss the supported configuration argument and environment variable even though both are documented elsewhere.

### Interface changes

None.

### Behavioral changes

- The missing-cluster error now points users to `Config(cluster_id="...")` and `DATABRICKS_CLUSTER_ID`.

### Internal changes

- Updated the existing DBUtils regression test to assert the actionable guidance.

## How is this tested?

- `uv run pytest tests/test_dbutils.py -q` (19 passed)
- `uv run ruff format --check databricks/sdk/dbutils.py tests/test_dbutils.py`
- `git diff --check`


TangoEnSkai:
Updated this branch with upstream main in 0ab3e56824affa5d0fbef23fdde83882efb6586d. The PR still changes only the missing-cluster guidance and its regression assertion.

Revalidated locally: all 19 tests in tests/test_dbutils.py pass, Ruff format/check pass for both changed files, and the diff against upstream main passes git diff --check. No workspace integration run was performed.

Could a maintainer review and authorize the fork workflows for this head? I understand from CONTRIBUTING.md that an accepted external change is reapplied internally rather than merged directly here.

